PR #70, opened 2017-05-05
Pretify targets json

martinstarman:
Targets JSON from ```--debug``` option is hard to read. With this change it will look like
```
[debug] Targets:
  [
  {
    "nodeRange": "node7",
    "platform": "linux",
    "arch": "x64",
    "output": "...",
    "forceBuild": false,
    "fabricator": {
      "nodeRange": "node7",
      "platform": "win",
      "arch": "x64",
      "forceBuild": false,
      "binaryPath": "..."
    },
    "binaryPath": "...",
    "slash": "/"
  },
  {
    "nodeRange": "node7",
    "platform": "macos",
    "arch": "x64",
    "output": "...",
    "forceBuild": false,
    "fabricator": {
      "nodeRange": "node7",
      "platform": "win",
      "arch": "x64",
      "forceBuild": false,
      "binaryPath": "..."
    },
    "binaryPath": "...",
    "slash": "/"
  }
]
```


igorklopov:
Prettifying is a good idea, but removing a check is not. Pkg should stop i case of `-t node6 -t node7` in favour of `-t node6,node7`. Probably the error message is worth changing thou.

martinstarman:
Youre right! I tried several ```--target``` inputs but this I missed. Reverted.

igorklopov:
Ok. Let's try.

martinstarman:
Ok cool :) I will send another PR.

PR #77, opened 2017-05-06
Show version before compilation

### LINKED ISSUE #36: Show pkg version before compilation

martinstarman:
My attempt to close #36.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,6 +9,7 @@ import packer from './packer.js';
 import path from 'path';
 import { plusx } from './chmod.js';
 import producer from './producer.js';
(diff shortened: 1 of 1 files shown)


igorklopov:
Oh, it will not show version when targets are specified. Also ```log.info(`pkg@${version}`);```

martinstarman:
Oh, sorry. I was fixing lint issues and ends with old version.

igorklopov:
You know, let's place that log.info right after your '-v' handler. This way we have version logging parts close to each other.

martinstarman:
Done!

PR #92, opened 2017-05-10
Support for multiple -t arguments

martinstarman:
Support multiple `-t` arguments which we discuss in #70. Im not sure if it should be in `--help`. I think string is preferred way.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -260,12 +260,11 @@ export async function exec (argv2) { // eslint-disable-line complexity
   // targets
 
   const sTargets = argv.t || argv.target || argv.targets || '';
-  if (typeof sTargets !== 'string') {
-    throw wasReported(`Something is wrong near ${JSON.stringify(sTargets)}`);
-  }
 
   let targets = parseTargets(
-    sTargets.split(',').filter((t) => t)
+    Array.isArray(sTargets)
+      ? sTargets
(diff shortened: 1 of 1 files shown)


lib/index.js
```diff
@@ -260,12 +260,11 @@ export async function exec (argv2) { // eslint-disable-line complexity
...
   const sTargets = argv.t || argv.target || argv.targets || '';
-  if (typeof sTargets !== 'string') {
-    throw wasReported(`Something is wrong near ${JSON.stringify(sTargets)}`);
-  }
 
   let targets = parseTargets(
-    sTargets.split(',').filter((t) => t)
+    Array.isArray(sTargets)
+      ? sTargets
+      : sTargets.split(',').filter((t) => t)
```
cvan: do you want to trim any whitespace?

martinstarman: This is the current code. I think this is ok, or maybe in next PR.

Maskime: Maybe not my place to say it. But have you considered working with [command-line-args](https://github.com/75lb/command-line-args). It can handle such things as multiple values for commands arguments and it has a brother project [command-line-usage](https://github.com/75lb/command-line-usage) that can generate the help documentation. It was a real time saver on my last project can't recommend it enough.

rauchg:
We want to switch `pkg` to using https://github.com/zeit/arg

PR #154, opened 2017-06-30
dictionary: bunyan

weirongxu:
fix bunyan dynamic require

### DIFF
--- a/dictionary/bunyan.js
+++ b/dictionary/bunyan.js
@@ -1,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  pkg: {
+    patches: {
+      'lib/bunyan.js': [
+        'mv = require(\'mv\' + \'\');',
+        'mv = require(\'mv\');'
+      ]
+    }
+  }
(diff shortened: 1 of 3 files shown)


igorklopov:
Is it only to remove a warnings? Does it break at run time?
Try as `require('mv', 'may-exclude');`

weirongxu:
@igorklopov Its break at run time when use `rotating-file` steam.
```js
const logger = require('bunyan').createLogger({
  name: 'pkg',
  streams: [{
    type: 'rotating-file',
    path: 'pkg.log',
    level: 'info',
  }]
})
```

weirongxu:
@igorklopov I try to add `require('mv', 'may-exclude')`, Its work very well.

igorklopov:
`require('mv', 'may-exclude');` still absent in changes. Add it please

igorklopov:
Oh, sorry, now i get the problem. Please undo `may-exclude`. It is not needed really. And then i merge 👍

PR #203, opened 2017-08-22
add --bundledir param so that you can debug what files get bundled

FabledWeb:
Sometimes you might want to debug (or export) the files that actually get put into the binary snapshot. This adds a flag to let you do just that.

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -16,6 +16,8 @@ export default function help () {
     -d, --debug      show more information during packaging process [off]
     -b, --build      don't download prebuilt base binaries, build them
     --public         speed up and disclose the sources of top-level project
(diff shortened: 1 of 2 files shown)


igorklopov:
I really appreciate your help, but i have another plan for this feature. Particularly, i plan to implement json output. So that `--json` and `--debug` together will give you the list of files. Another project (mine or community-driven) parses the json and copies files as you describe.

FabledWeb:
@igorklopov That seems like a reasonable plan. I'll just use my fork for now, so feel free to close/reject this PR if you want. Also feel free to reuse any of the above code in the simple "other project" that can parse and copy the files if you find it helpful.

Root-Core:
I would absolutely love some feature like this! Is there an alternative to this pull request yet?

PR #244, opened 2017-09-25
Include version number in output file name

pascalbayer:
This PR introduces the version number in the output file name.

Including the version number has benefits for versioning and releases. It could be considered to make the version number configurable via CLI option.


igorklopov:
Are you sure you want to include the version of pkg, not the version of your software?

pascalbayer:
I intended to include the version of my own binary, if that is something you want to accept as a PR, I'll update my code

PR #351, opened 2018-02-12
Add permission flag to allow private modules

szaffarano:
Add --permissive flag that allow private and unlicensed modules be
processed when using --no-bytecode flag.  For more information
please see zeit/pkg#145


igorklopov:
Implemented it in a little different way in https://github.com/zeit/pkg/commit/1919967ad178f4ee8eb0253f3b7fdd34f7771003. Thanks! Closing

PR #412, opened 2018-05-03
Define fs.realpath.native

corevo:
Node 9.2 specified [fs.realpath.native](https://nodejs.org/api/fs.html#fs_fs_realpath_native_path_options_callback),  
I'm trying to pkg a dependency that depends on it being defined.  
The problem is that it's using [realpath](http://man7.org/linux/man-pages/man3/realpath.3.html) under the hood, which will not work for the snapshot fs, so I'm redirecting to the non-native one.  

Perhaps a better solution is to check wether the path is in the snapshot or not, and if not use the native one.


acdibble:
@igorklopov, is it possible to get this or another fix merged in? `fs.realpath.native` and `fs.realpathSync.native` aren't currently usable with pkg on any LTS releases of Node.js.

leerob:
Closing as stale - thank you anyway 🙏

PR #451, opened 2018-06-22
Fix eslint on Windows

shawwn:
This PR disables the `linebreak-style` eslint rule.


igorklopov:
Let's leave unix styles, even nn Windows

PR #461, opened 2018-07-05
Support Node 10 when using the internal fs module real functions

bergheim:
As per https://github.com/nodejs/node/issues/17076 the functions have been
renamed in node now. This reenables dynamic loading of modules using the node10
runtime.


Torsten85:
Same as https://github.com/zeit/pkg/pull/452 ;)

unitof:
Until this merges I've just gone ahead an `npm i`’d `pkg@bergheim/pkg#fs-internalmodule-rename`.

igorklopov:
Fixed! Thank you!

PR #462, opened 2018-07-06
Compatibility with olders versions of Node (Error fixed).

Davd96:
Fixes an error occurred when trying to compile projects with old versions of Node that do not support the Buffer library.


rauchg:
We don't intend to support old versions of Node wihth no `Buffer`

PR #538, opened 2018-09-25
Fix handling of files with for-await-of loops

mattwiller:
Trying to run `pkg` on a file with a for-await-of loops currently throws an error complaining that a parser plugin is needed.  This adds the plugin and a test to ensure it works correctly.


mattwiller:
I'm not sure why some of the tests are failing — they seem to be having issues in unrelated npm tests.

aviflax:
Any chance this change could make it into a release anytime soon? Would be nice…

rauchg:
This sounds great to me. Merging.

PR #579, opened 2018-11-30
Use internalModuleReadJSON where it is defined.

pavelfeldman:
Fixes https://github.com/zeit/pkg/issues/467


igorklopov:
Thanks!

PR #580, opened 2018-11-30
Added support for async generators

leo:
We need this to make [this](https://github.com/zeit/now-cli/pull/1710) work by adding support for async generators.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "babel": {
     "plugins": [
       "transform-async-to-generator",
+      "transform-regenerator",
       "transform-runtime"
     ],
     "presets": [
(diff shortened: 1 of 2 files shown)


leo:
As seen in https://github.com/zeit/now-cli/pull/1710, it fixes our problem (I published `pkg@4.3.4-canary.1` in order to test it out).

timneutkens:
Can we add a regression test for this?

timneutkens:
As per talking with @leo:

Node.js supports generator syntax:

```
function* a() {
  yield 1;
}
```

But it doesn't support `async` generator syntax:

```
async function* a() {
  yield 1;
}
```

leo:
Closing in favor of https://github.com/zeit/now-cli/pull/1710/commits/158f96287ebdc7fd5cf0671ca451e0649c96b5de

PR #606, opened 2019-01-09
Update globby to fix broken dependency.

randallknutson:


randallknutson:
https://github.com/zeit/pkg/issues/603

PR #610, opened 2019-01-10
Update dependencies and remove fixed versioning

robcresswell:
(Same intent as https://github.com/zeit/pkg-fetch/pull/54)

This patch updates some deps with non-major version changes, and also
relaxes the versioning used in the package.json to allow bugfix
releases.

Using pinned versions is probably a well intentioned but naive
restriction that prevents a consumer from automatically receiving
security fixes unless the package author approves them, which can take a
long time. Pinned dependencies don't work in Node-land anyway, as NPM
will happily give you multiple copies of a specific library at different
versions, and pins do also not prevent transitive dependencies from
including version ranges anyway. For these reasons, it is more a
security risk than a reliability insurance.


robcresswell:
This should hopefully avoid some of the issues like `globby` in the future.

igorklopov:
Let's leave `resolve` package pinned to `1.6.0` for now.

robcresswell:
Hey @igorklopov, I've updated it to keep resolve at `~1.6.0`, is that okay?

robcresswell:
Hey again @igorklopov any work needed here? I'll close it otherwise if its not wanted :innocent:

robcresswell:
I don't want to keep eyes on this forever. If it is useful please restore / message me

PR #622, opened 2019-01-22
Feature Preprocessor

DanielSchuech:
Hello,

The idea is to allow the usage of preprocessors similar to webpack. There you can use different loaders to preprocess your files.

### General
With this PR you can define preprocessors in the pkg config in the package json. This can be used e.g. to compile your es6, typescript or flow code before it gets bundled into the binary. The advantage is that you dont need any precompile step.

### Advantages for Lerna Mono Repos
You dont have to compile every package in a lerna mono repo. This packages can just link to the es6 (or flow or typescript) code.

That is just one use case for a preprocessor. There are more opportunities like minification.

Best regards
Daniel


leerob:
Hello! Thank you for contributing back, and sorry for the slow response. This is a really large change and unfortunately something outside the scope of what I'm trying to revive support for here. Ideally, I'd like to just clean things up and get newer versions of Node published.

Going to close this as stale 🙏 Please comment back if you'd like to take this to the finish line.

PR #656, opened 2019-04-02
Support BigInt for @babel/parser

danielgindi:


debovema:
Could we provide additional plugins with a configurable parameter instead ?

danielgindi:
@debovema That could be a separate PR, but this one is about compatibility with ES syntax which is already stage 3 and used in a lot of code. It *should* be provided "out of the box".

leerob:
Closing this as stale. Thanks anyway for your PR 🙏

PR #668, opened 2019-04-26
Fix pkg example express app

DaneTheory:
**The Issue**
Currently, following the instructions provided in the `README` result in `pkg` failing to build the executable. The error output references `node7` as being an undefined `target` in `package.json`.

**The Solution**
`pkg` relies on `pkg-fetch` for retrieving Node versions to be used as `targets` in building executables. As of `pkg-fetch` v2.5, no `node` v7.x.x is listed as an available source. Changing `node7` to `node10` fixes this issue as it is available via `pkg-fetch` v2.5.

**Additionally, this PR**

- Adds files/folders to .gitignore related to the example express app which should not be included in master.
- Improves README ducumentation for example express app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,13 @@
 
 # logs
 npm-debug.log
+
+# General macOS
+.DS_Store
+examples/.DS_Store
+
+# Example pkg'd application
+examples/express/express-example
+
+# Example dependencies
+examples/express/node_modules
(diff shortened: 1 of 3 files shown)


DaneTheory:
@igorklopov @leo Not sure what's causing the build errors. You both have contributed the most to this project. Any thoughts? Could be over looking something obvious.

igorklopov:
Thanks!

DaneTheory:
Np

PR #677, opened 2019-05-06
Allow to use any entrypoint

Luke265:
This is required for Node12 worker_threads module to work properly.

https://github.com/zeit/pkg-fetch/pull/65


igorklopov:
Thanks! But i think we must preserve checking it in main thread. So, closing it in favour of https://github.com/zeit/pkg/commit/589c6a5ee730bd56b4a263609b47896a9a5a06e3

PR #684, opened 2019-05-15
Added environment var section on Readme

robertsLando:


leerob (approved):
As we discussed, the README definitely needs a bit of work, but I see no harm merging this in as a base for showing other env options.

PR #720, opened 2019-06-17
Add missing stat.isSocket

zvin:


sommd:
Any reason this hasn't been merged yet? This fixes a problem I'm having with `tar-fs`, which is preventing me from packaging my app.

pdcastro:
@zvin and I are external contributors without permission to merge this.
@igorklopov might be able to...

pdcastro:
By the way, I can also confirm that this PR fixes a problem we were having with the [balena-cli](https://github.com/balena-io/balena-cli/). We've been successfully applying a patch over pkg 4.4.0 while this isn't merged:
https://github.com/balena-io/balena-cli/blob/v11.4.4/patches/pkg%2B4.4.0.patch

sommd:
It might also be work adding the missing [isBlockDevice](https://nodejs.org/api/fs.html#fs_stats_isblockdevice) and [isCharacterDevice](https://nodejs.org/api/fs.html#fs_stats_ischaracterdevice). I'm guessing they should always return false.

leerob (approved):
Thank you so much! Sorry for the slow response here.

PR #751, opened 2019-08-10
add support for `open` (renamed from `opn`)

sergioramos:


sindresorhus:
It would make more sense to do the opposite: Rename `opn` to `open` and then add a stub for `opn` that just imports `open`.

igorklopov:
Both `open` and `opn` are supported since https://github.com/zeit/pkg/commit/db8487acd2d40af7b0cb1e63e133a707f93d725f, https://github.com/zeit/pkg/commit/fad82c5885d9518a73b901efd5d31f752f35df35 and https://github.com/zeit/pkg/commit/82737779d592842b4fa35c84c425cc0ca19280c5. Thanks!

PR #754, opened 2019-08-15
allow option full path(walker.js)

andepzai:
Just a small change, support to use path.join outside pkg
useful to use exact path in config file

### DIFF
--- a/lib/walker.js
+++ b/lib/walker.js
@@ -65,7 +65,7 @@ function upon (p, base) {
     p = p.slice(1);
     negate = true;
   }
-  p = path.join(base, p);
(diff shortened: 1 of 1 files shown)


igorklopov:
I'll try adding `isAbsolute` soon

igorklopov:
If `HOME` is `/root` (for example in CI), then a malicious package can take `/root/.ssh` into executable so that private keys will leak as assets. And assets are extractable from the executable. So, let's mark it "wont fix" for now until i come up with a better idea.

leerob:
Closing as won't fix based on above conversation 🙏

PR #791, opened 2019-11-04
Update node-forge pkg scripts in dictionary

SlyryD:
- Update node-forge pkg scripts in dictionary ("js" directory is now "lib" in node-forge package).
- Breaking change for those who are using old node-forge packages. Is there another way to do this?


igorklopov:
Fixed in https://github.com/zeit/pkg/commit/466b5fda8464de6c4363af27cd796fe2babcf828. Thanks!

PR #823, opened 2019-12-31
Add ZEIT Now Disclaimer

msweeneydev:
This PR adds a disclaimer to the README, advising users that `pkg` is not required to use ZEIT Now.


msweeneydev:
Correct, added codeblock usage to be consistent with the rest of the README also.

PR #853, opened 2020-02-21
Load image over HTTPs

charliea21:
This change should stop the NPM site from displaying 'Not Secure' due to an image being loaded over HTTP, not HTTPs.


igorklopov:
Thanks!

PR #862, opened 2020-03-11
Enhanced example

pierrejochem:
I added some enhancements to the example


leerob:
Thank you for your contribution! We're going to keep the example as-is for now, with larger improvements needed.

PR #868, opened 2020-03-16
Added support for user-defined pkg config for 3rd party dependencies

pkt-zer0:
The title pretty much says it all. pkg itself has a `dictionary` object which can define the config used for a given package, this just adds user-defined entries to it.

Previous alternative solutions involve defining assets as `node_modules/some_package/assets/*`... which works as long as that's actually where that package gets installed. But with NPM's package hoisting/deduplication, that's not necessarily a fixed location. So this allows end users to do exactly what they want: add the missing asset/script configuration to a 3rd party package.

I didn't update the docs yet, I'm not sure if this is the proper approach to begin with. (Though it does already solve the problem I had with packaging pdfkit and others)


igorklopov:
Great idea! I suggest calling it `dictionaries` to align with current naming. Wdyt?

pkt-zer0:
I'm fine with either name. `dictionaries` seemed a bit non-descript to me, but maybe with more context it's obvious. I picked `externals` because this is the part where you can configure pkg-related settings not for your _own_ package, but your external dependencies.

Anyway, let me know what you prefer. In the meantime I'll update the README in a new PR.

Issue #920, opened 2020-06-01
Issues caused by .mjs (esm modules) in dependencies


alexk111:
Getting `Failed to make bytecode node12-x64 for file ...` errors when building a package. I configured it to compile all modules as-is (assets) and only bytecode MY-MODULE which is a commonjs module and shouldn't have any issues by the following pkg settings:
```js
pkg: {
    "assets": [
      "node_modules/**/*",
      "!node_modules/MY-MODULE/**/*.js"
    ],
    "scripts": [
      "node_modules/MY-MODULE/**/*.js"
    ]
}
```

Everything works as expected, excepting one thing. It keeps trying to make bytecode for dependencies specified in one of MY-MODULE dependency's package.json. This causes the error.

Did some research on the error and here is what I've found.
The error is thrown by Fabricator's `script`. It can only bytecode commonjs modules, due to the use of `module.wrap`. And since one of those dependencies is a esm module (.mjs file) the module wrapping applied to it produces an invalid code.

Next I discovered that pkg has a `dictionary` directory which allows to overwrite properties in package.json files of any dependencies for packaging purposes. This allowed me to remove the unneeded dependencies with:
```
'use strict';

module.exports = {
  dependencies: undefined
};
```
And this solved the issue. All the dependencies are in the built file (as per the ```{"assets": "node_modules/**/*"}``` setting in the root package.json) and pkg doesn't try to bytecode the ones removed by the ```{ dependencies: undefined }```. 

An ideal solution would be to prevent pkg from making bytecode for dependency files covered by the assets glob. Or if it will break something (I don't see the whole picture here :), then being able to do that manually via the dictionary feature is also should be ok. But this would require some way to do that on the user side, instead of updating pkg's dictionary.

There is a [commit](https://github.com/vercel/pkg/commit/2ad45486ddf3ba3642623739adbd03a47294f789) which implements that (if I understand that right), however it's not released yet. @igorklopov any ETA on this?

alexk111:
Sent a pull request enabling to overwrite `dependencies` properties of external modules from the root package.json.

Example for the above-mentioned scenario, when you need to disable bytecoding for ESM-MODULE from `dependencies` list of some MODULE your package depends on:
```
"pkg": {
    "assets": [
      "node_modules/**/*",
      "!node_modules/MY-MODULE/**/*.js"
    ],
    "scripts": [
      "node_modules/MY-MODULE/**/*.js"
    ],
    "dictionary": {
      "MODULE": {
[...]

PR #922, opened 2020-06-01
Set dictionary objects as-is

alexk111:
Since the feature was not publicly released yet, I thought that it will be safe to modify the assignment of dictionary elements instead of implementing a separate property. This change will enable to update other package.json properties, such as `dependencies`, to overcome issues described in #920


alexk111:
Also commited a fix for a bug causing `dependencies` removal from the dictionary store, thus making it unavailable on 2nd access.

PR #929, opened 2020-06-05
add --always-compile option to compile even public packages

ryanroe:
background:
In case of those users who deployed internal npm servers, their core bussiness logic may stay in `node_modules` directory, so they might  want to compile js files in `node_modules` to protect their sources.

and I think this may be a nice feature , so I add an option for cli users .


ryanroe:
seems can be done by putting node_modules in `scripts` field , so closed :)

saikoux12:
哥们，我也遇到类似的问题，方不方便加个联系方式

saikoux12:
按照你这样改。打包完后，依然可以在可执行文件内看到源码

Issue #1019, opened 2020-12-28
Using PKG with Webpack => why /snapshot/project/node_modules ?


panpansh:
I have no error all is working fine.

Looking inside /snapshot/project/ : I don't know why there is some node_modules here.

would there be a possibility not to include any node_modules please ? (because using webpack).

Wishing you happy holidays <3

shinylightdev:
You may want to post your package.json and how you're calling pkg.

panpansh:
```js
"scripts": {
    ...
    "win:pkg": "cross-env NODE_ENV=production pkg --targets=win ./bundle.js --output ./dist/client.exe --config ./package.json",
    ...
},
"pkg": {
    "assets": [
         "assets/**/*"
    ]
}
```
is there a simple way to force to not import any node_modules ?

panpansh:
Ok so, I just need to move my bundle.js inside my dist folder by example. and dont let him next to node_modules.
Thats all.
Regards

PR #1049, opened 2021-02-22
add --no-derivatives

fatWill:
#### background:
I need to extract all the contents of `node_modules`, set the `NODE_PATH` environment variable, and read the external `node_modules` through the dependencies of `node`, so that my binary files will not be large

so I plan to use an option to set whether to complete the packaging based on the `assets` and `script` set in the `pkg config`without automatically looking for dependencies.

such as this issue(#1019)

### DIFF
--- a/lib/help.js
+++ b/lib/help.js
@@ -6,16 +6,17 @@ export default function help () {
 
   ${chalk.dim('Options:')}
 
-    -h, --help       output usage information
-    -v, --version    output pkg version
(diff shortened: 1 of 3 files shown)


fatWill:
have anybody?

leerob:
At this time, I don't we should add this. If there are multiple people interested in this via an existing or new issue (thumbs up), we could reconsider. If it's blocking, you can use `patch-package` to add this in your application. 

Thanks!

PR #1073, opened 2021-03-22
add workflow for GitHub Actions

erossignon:


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+      matrix:
+        node-version: [10.x, 14.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
```
robertsLando, line 36: I would replace all these lines with a single `run: |`, also did you try to keep only the npm install and npm test commands?

robertsLando (changes requested):

robertsLando:
Also remember to replace the badge in readme with the workflow badge

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+
+  build_on_windows:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
```
robertsLando, line 20: Here I see a matrix but there is no one specified in this job

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+    - run: npm install
+    - run: npm test
+
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 14.x]
```
robertsLando, line 28: Why don't include 12.x too?

robertsLando (changes requested):

.github/workflows/workflow.yml
jesec: Maybe consolidate them into `Install dependencies` step. 

`npm install` and `npm test` should remain separate steps.

.github/workflows/workflow.yml
jesec: Yes. And, add Current (`15.x`) as well. It is important for this project to track the latest developments of Node.js.

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+    - run: sudo apt-get update -qq
+    - run: sudo apt-get -yq --no-install-suggests --no-install-recommends --force-yes install gcc-4.8 g++-4.8
+    - run: sudo apt-get -yq --no-install-suggests --no-install-recommends --force-yes install libavahi-compat-libdnssd-dev # npm install mdns
+    - run: export CXX="g++-4.8" CC="gcc-4.8"
```
jesec, line 40: GCC 4.8? Not acceptable in 2021. 

Why not simply use `build-essential`?

I would also like to see a workflow for the latest GCC and Clang.**

** I can do this myself later, though.

leerob:
Thanks for this! Looking great - agree with the feedback above 🚀

robertsLando:
Supersided by #1074

erossignon:
Thank you guy, let's close this one.

PR #1076, opened 2021-03-24
set up and run prettier

hipstersmoothie:
Prettier has become the standard for JavaScript formatting. Adding it to the repo makes contribution and PR review easier.


jesec:
I agree that we should use prettier.

But we have to deal with ESLint first.

robertsLando:
We should firstly remove the eslint-klopov and use standard eslints

leerob:
I have _opinions_:

- Single quotes
- No semicolons (can be persuaded)

I'm fine with whatever order you prefer - Prettier first, then ESLint, or vice verse.

hipstersmoothie:
I prefer the prettier defaults, but that's only because I don't care too much. What does everyone here think? Should we go with the default @leerob suggested?

hipstersmoothie:
I'll fix/update the eslint configs as a follow up PR

hipstersmoothie:
I've switched the prettier defaults to `singleQuote: true` to match the format of the current code. Some tests actually have the semi baked in to some of the assertions.

hipstersmoothie:
Think this one is good to merge @robertsLando

hipstersmoothie:
I think after the prettier and eslint work is merged next steps I'd like to take are:

1. TypeScript conversion
2. Enforce jsDoc on all functions

leerob (approved):

leerob:
Do I need to turn off some of these checks? Which ones do we need? Just the top 3?

hipstersmoothie:
@leerob I think I saw @robertsLando commend on another thread that those 3 are the only required ones at the moment

PR #1062, opened 2021-03-13
Update all dependencies to latest.

erossignon:

PR #1100, opened 2021-03-26
chore: make pkg-fetch dep static

robertsLando:


erossignon:
This fix is required for #1062 to work

erossignon (approved):

erossignon (approved):

leerob (approved):

leerob:
This is just temporary, right? Until we fix the issue that was introduced?

robertsLando:
Humm nope I would keep this anyway. Every new release of pkg-fetch needs to firstly have the binaries builded and available then it can be bumped here

PR #1105, opened 2021-03-26
add Windows build on CI

erossignon:


robertsLando (approved):
Yep I removed it just to make tests fasters as seems that windows builds takes twice the time of linux ones. BTW we could keep them there and just make the linux ones required.

PR #1106, opened 2021-03-29
Add section about bytecode flag to README.md

whexberg:
I admit it's a bit wordy. But I figured this could be a big trade off for some users, so it may need some explanation.


robertsLando (approved):
LGTM

PR #1107, opened 2021-03-29
extend linter to test files

erossignon:


.eslintrc
```diff
@@ -20,10 +21,41 @@
...
+        "import/no-dynamic-require": "off",
+        "import/no-extraneous-dependencies": "off",
+        "import/newline-after-import": "off",
+        "import/no-unresolved": "off",
+        "import/no-useless-path-segments": "off",
+        "import/order": "off",
+        "no-console": "off",
+        "no-lonely-if": "off",
+        "no-multi-assign": "off",
+        "no-undef": "off",
```
robertsLando, line 48: Could we remove some rules here?

robertsLando (changes requested):

.eslintrc
erossignon: the PR will be massive so many tests file will need updating. Are you ready ?

hipstersmoothie: We could just take this PR as is and have follow up PRs that removes flags (I could help)

robertsLando: Ok for me, just asking :) Typescript is good but if we disable strict and other rules it becomes almost useless :laughing:

robertsLando: We could disable one by one

robertsLando (approved):

PR #1111, opened 2021-03-31
WIP: developers docs

robertsLando:
@erossignon @hipstersmoothie @jesec @leerob when you have time please help me with this :)

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -1,0 +1,128 @@
+# Developers
+
+This file is used to help new developers understanding how pkg works at low level.
+
+## Setup
+
+```bash
+git clone https://github.com/vercel/pkg.git
+cd pkg
+# install deps and build project (see `prepare` script)
+yarn install
+# link pkg globally
+yarn link
+# link pkg inside your app
+cd ../my-test-app
+yarn link pkg
+```
+
+With this setup you will be able to test your customized version of pkg inside your application.
+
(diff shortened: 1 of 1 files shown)


robertsLando:
@leafac I have used some things I found in your caxa docs as it was very well explained

leafac:
I’m glad that you found it helpful.

jesec:
Good job. Though, I think we can instead use the "Wiki" of Github for better flexibility, and them occasionally reorganize and commit them to the project repo itself.

jesec:
> @jesec I think [docsify](https://docsify.js.org/#/) is also a good alternative, I personally don't like github wiki. I use docsify in other my projets and the result is so good. BTW For now I will focus on adding things here and then we can switch to docsify in another PR, I can do that

It is very easy to edit Github Wiki. Other options are not that flexible, as they would require commit, PR, review, merge, etc.

We can of course use whatever doc generator for the final and official doc, but I think Github Wiki is the most flexible one option for an evolving documentation.

leerob:
Let's start with GitHub Wiki and when it grows large enough, I can create a docs site for this project!

leerob:
Created https://github.com/vercel/pkg/wiki

robertsLando:
Ok for the wiki,  should we move the content of this pr there and close this so?

leerob:
Sounds good 👍

Issue #50, opened 2017-05-01
Possibility to reduce packages size by 4x


pi0:
Hi! Just done a quick test on windows builds. Bundle size can be reduced from **24,964 KB** to **6,646 KB** (3.75x smaller) with doing simple *xz* compression. This may even help more when having more npm dependencies. Adding this option may be super useful for small and embedded CLI/Webservers :) I've just opened this discussion before going further to know others idea on this.

**UPDATE**: To be clear, this topic and what i am talking about is *SFX* packages not js minification & optimizing.

pascalandy:
@pi0 Do you see any potential impact on the performances of you Node App?

I'm considering this for my Ghost installation. Cheers!

roccomuso:
@pi0 How did you do that compression? could you give an example?

jgoux:
Hi,

as we're talking about bundle size, does `pkg` apply some kind of optimization / tree shaking / minification internally?

If I apply something like prepack + uglify before packaging, will my final packaged app be lighter?

anmonteiro:
running `strip` on the generated binary should also help on UNIX platforms

igorklopov:
The idea behind pkg is to introduce minimal changes over original nodejs. Btw, if you run `strip` you can lose the ability to use native addons (`.node` files).

pi0:
Hi. Sorry answering late :)

@pascalandy About performance impact: I think if we do extraction on startup, it may only affect startup time and not run-time itself.

@roccomuso My first test was just using tar! But it would be easy adding self-extract bits on binary's head.(exe inside exe)

@anmonteiro  About js minification, pkg is absolutely not webpack! And i don't think this would help a lot, main size suffer is from Node.js binary itself. And we could just make a blind compression on the whole things. (Also @igorklopov is correct about `.node` files)

pi0:
Some interesting projects we could get idea:

- https://github.com/megastep/makeself
- [more coming to list]

lenovouser:
@anmonteiro `strip` doesn't really work. It reduces the binary size but the app doesn't run anymore. You get the interactive node console 😆 

![](https://i.lenovouser.me/fmMKTLI4.jpg)

That is very weird. Why is this happening?

igorklopov:
`strip` strips payload that contains your project packed. Without that payload the binary behaves like original node.js binary.

anmonteiro:
[...]

PR #1115, opened 2021-03-31
compress files in virtual file system (#885)

### LINKED ISSUE #885: Zip executables ?
Can we have an option like  `--zip` to compress executables as soon as they are built 🤔

erossignon:
Fixes #885


hipstersmoothie:
Would be cool to see some before and after data on this

erossignon:
This PR enables compression of Virtual File System table and Zip/Brotli compression of each individual file content.
This allows a gain of up to 60% of the embedded file system.


```` 
node.exe size  = 71917203
No compression =  175703
        Δ GZip =  -100540 ( -57 %)
      Δ Brotli =  -109057 ( -62 %)
````

jesec:
If possible, IMO we can try to make it possible to compress the entire binary, not only the virtual file system.

robertsLando:
@erossignon Nice work here! BTW I have linked the pr that was asking for zip executables, what do you think about them? I know https://www.npmjs.com/package/caxa works in this way for example

erossignon:
> If possible, IMO we can try to make it possible to compress the entire binary, not only the virtual file system.

Yes, The executable that we use from pkg_fetch is far too big, It should be compiled with minimum ICU to significantly reduce its size. then we could had some options in pkg to reference or add the ICU as a external module ( optional)

I have no idea how to compress the entire executable further... if you have a idea, let me know.

robertsLando:
@erossignon caxa does this, more info in the repo: https://github.com/leafac/caxa#readme

robertsLando (approved):
LGTM.

cc @leerob @jesec

erossignon:
@leerob @jesec , do you have any objection that would prevent merging  ?

leerob:
I'd like to wait until we cut a release of pkg that's fetching from pkg-fetch properly.

erossignon:
> I'd like to wait until we cut a release of pkg that's fetching from pkg-fetch properly.

OK, I'll keep rebasing and keep the code up to date until new version of pkg-fetch is fully integrated

Hypfer:
> I have no idea how to compress the entire executable further... if you have a idea, let me know.

[UPX](https://github.com/upx/upx) might be an option

See:
https://github.com/vercel/pkg/issues/50#issuecomment-789951164

Relevant quotes

> https://sourceforge.net/p/upx/discussion/6806/thread/79e2a6b8/

> As it is now, the `readPrelude` function patched into the nodejs binary will open `/proc/self/exe` and try reading the overlay from the offset replaced into the binary at packaging time. However since the binary found at `/proc/self/exe` is the compressed version, all those offsets are of course just wrong and even if they weren't they would still read compressed data.

> To get pkg to work with upx, one approach could be to update the `readPrelude` function so that it uses relative offsets from the end of the file and extend the pkg process so that it first bakes in the commandline options, then packs that binary and finally appends the rest of the code.

erossignon:
@leerob  can we incorporated to a 5.1  release ?

leerob:
5.1 is out, but we can get this in 5.2! 😁

alitas:
On an unrelated note, there seems to be a typo on a few lines in the commits, both in code and documentation, using Brolti instead of Brotli. It might cause some confusion among users.

robertsLando:
@erossignon Could you fix the typos?

erossignon:
let's go

Hypfer:
When testing this + upx, my webserver spontaneusly started to deliver compressed assets
image

This may or may not be a side-effect of the upx compression though

robertsLando:
@jesec @leerob Could we merge this?

erossignon:
>  It did work fine initially and only stopped working after a few minutes of runtime

This could mean  that the size of the executable could changed at runtime somehow ?! your technique for relative offset, reads the executable size each time to find the start of the prelude by subtracting the payload size. Strange.

PR #1120, opened 2021-04-06
Support mkdir at mountpoints

phated:
I'm working on a CLI tool that bundles some code, but then mirrors the directory structure out to the filesystem. This is done using mountpoints and this patch that allows `fs.mkdir` to apply to them.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -453,6 +453,8 @@ function payloadFileSync(pointer) {
   ancestor.exists = fs.exists;
   ancestor.accessSync = fs.accessSync;
   ancestor.access = fs.access;
+  ancestor.mkdirSync = fs.mkdirSync;
+  ancestor.mkdir = fs.mkdir;
 
   var windows = process.platform === 'win32';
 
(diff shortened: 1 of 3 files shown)


prelude/bootstrap.js
```diff
@@ -1241,6 +1243,40 @@ function payloadFileSync(pointer) {
...
   };
 
+  // ///////////////////////////////////////////////////////////////
+  // mkdir /////////////////////////////////////////////////////////
+  // ///////////////////////////////////////////////////////////////
+
+  function mkdirInSnapshot(path_, cb) {
+    var cb2 = cb || rethrow;
+    return cb2(
+      new Error('Cannot mkdir in a snapshot. Try mountpoints instead.')
```
phated, line 1253: Is there a better error for here?

robertsLando (commented):
Could you add a test for this?

prelude/bootstrap.js
robertsLando: I would instead change the name of the method in something more user friendly, something like `mkdirThrowInSnapshot`

phated: I went with `mkdirFailInSnapshot` because it doesn't throw if a callback is provided to the async method, but it still fails.

phated:
Test added! I named it with the PR number. Let me know if it should be something different.

phated:
Hmm, this had the same pnpm failure as #1121

robertsLando:
Seems that the problem is node version, strange them were not failing before... @erossignon could you fix those tests polease?

erossignon:
Let me provide a patch for the pnpm issue... 
We will skip the pnpm test when nodejs 10  is detected.

erossignon:
please merge #1122 first

robertsLando (approved):
LGTM

robertsLando:
This can be merged once check pass

robertsLando:
@leerob can we merge this?

PR #1122, opened 2021-04-07
Skip pnpm tests when nodeversion is < 12

erossignon:


robertsLando (approved):
In future we could also check supported nodejs versions in package.json and skip tests automatically if not there. BTW LGTM for now

PR #1140, opened 2021-04-20
Fix isPublic check for licenses array

### LINKED ISSUE #1139: --no-bytecode is currently broken
At some point, `--no-bytecode` broke.
Unfortunately, I cannot pinpoint exactly when that happend, however what I can say is that it did work at some point, since I've been using it for the last three(?) years

A very basic example to reproduce this is to use the hello world code of expressjs

<details>
<summary>package.json</summary>

```json
{
  "name": "pkg_bytecode_broken_test",
  "version": "1.0.0",
  "main": "app.js",
  "bin": "app.js",
  "dependencies": {
    "express": "4.17.1",
    "pkg": "git+https://github.com/vercel/pkg.git#4.5.1"
  }
}
```

</details>

<details>
<summary>app.js</summary>

```js
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
    res.send('Hello World!')
})

app.listen(port, () => {
    console.log(`Example app listening at http://localhost:${port}`)
})

```

</details>

As you can see, it doesn't work

```
PS D:\WebstormProjects\pkg_bytecode_broken_test> pkg --target win --no-bytecode .
> pkg@4.5.1
> Error! --no-bytecode and no source breaks final executable
  D:\WebstormProjects\pkg_bytecode_broken_test\node_modules\uid2\index.js
  Please run with "-d" and without "--no-bytecode" first, and make
  sure that debug log does not contain "was included as bytecode".
```

Interestingly, I can get it to build successfully by editing this code so that it sets `record[STORE_CONTENT]` to true if the record does have a body instead of throwing. That is of course a hack and may cause problems later on, however the resulting binary does seem to work just fine at first glance.
https://github.com/vercel/pkg/blob/63f4ce03f71f926521eb8f94dd49b24fb58190ed/lib/packer.ts#L86-L97

iirc I had a similar issue a while ago and it was magically fixed by installing the dependencies via yarn instead of npm. Unfortunately, this doesn't seem to help now though.

Anways, I'm fairly certain that it's a bug in the walker. Symlinks? Issues with case-sensitive stuff? Something like that maybe

Hypfer:
As it turns out, not including bytecode is broken for modules that have a licenses array in their package.json.

One example of this is `utils-merge` which is a dependency of `express`.

Without this check, `license` will become undefined since there's no `type` property on that array.
This causes the module to be listed as non-public which means that it doesn't get included as CONTENT
That in turn triggers the exception when building without bytecode.

Anyways, this was not fun to debug, because there's no mention anywhere of this isPublic check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ into executable), it may be useful to look through the log.
 By default, your source code is precompiled to v8 bytecode before being written
 to the output file. To disable this feature, pass `--no-bytecode` to `pkg`.
 
-> Why would you want to do this?
+#### Why would you want to do this?
 
 If you need a reproducible build
 process where your executable hashes (e.g. md5, sha1, sha256, etc.) are the
@@ -157,7 +157,7 @@ same value between builds. Because compiling bytecode is not deterministic
 results in executables with differing hashed values. Disabling bytecode
 compilation allows a given input to always have the same output.
 
-> Why would you NOT want to do this?
+#### Why would you NOT want to do this?
 
 While compiling to bytecode does not make your source code 100% secure, it does
 add a small layer of security/privacy/obscurity to your source code. Turning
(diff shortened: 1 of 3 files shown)


Hypfer:
Actually, this doesn't fix the issue #1139, because the module `uid2`, which is used in the example there, features no license at all and will therefore always be listed as private.

It doesn't matter for my main project and this still fixes a bug, however I suppose it would make sense to hide the `isPublic` check behind some kind of flag, so that it's just used by people that actually need it and not breaks foss projects for absolutely no reason

robertsLando (changes requested):
@Hypfer thanks for this fix, could you also mention this flag in docs please?

Hypfer:
Done

robertsLando (approved):
LGTM

PR #1144, opened 2021-04-23
Use "linuxstatic" to generate bytecode for different arch on Linux

jesec:
Currently `pkg` can't generate executable for different arch:

> Warning Failed to make bytecode node14-arm64 for file ...

as it attempts to run Node of different arch to make bytecode.

We can set up QEMU and binfmt to allow emulation. However, the "linux"
binary needs libraries from the system, and it is PITA to set up the
entire sysroot of a different arch.

This change workarounds the issue by using "linuxstatic" to generate
bytecode for different arch.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -154,9 +154,23 @@ function stringifyTargetForOutput(
   return a.join('-');
 }
 
-function fabricatorForTarget(target: NodeTarget) {
-  const { nodeRange, arch } = target;
-  return { nodeRange, platform: hostPlatform, arch };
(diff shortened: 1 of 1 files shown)


robertsLando (approved):
LGTM. Maybe adding a log when the `fabPlatform` is forced (explaining why) would be good

jesec:
> LGTM. Maybe adding a log when the `fabPlatform` is forced (explaining why) would be good

Later, we can add more documentation about compiling for different arch.

fcastilloec:
Will this change help compile for armv7? and probably help with #1148? If so, how can use this to generate an armv7 from an x86_64 arch?

Issue #909, opened 2020-05-21
child_process.fork with execArgv '--max-old-space-size' fail


RogerKang:
test.js:
```
const cp = require('child_process');
cp.fork('./forkChild.js', [], {
  execArgv: ['--max-old-space-size=2048'],
});
```
forkChild.js:
```
console.log('print from child');
```
and pkg.json:
```
{
  "assets": ["forkChild.js"]
}
```

And build with pkg:
`pkg -t node12-macos-x64 -o test -c pkg.json test.js`
then run:
`./test`
Exception shown:
```
internal/modules/cjs/loader.js:983
  throw err;
  ^

Error: Cannot find module '/xx/xx/pkgTest/--max-old-space-size=2048'
```

If we remove execArgv from test.js, it will be ok.
So I think this issue is caused by pkg's handling about child_process fork?

perminoveugenehuawei:
Same issue. Any workarounds?

perminoveugenehuawei:
I've found workaround - you can use worker_threads module and provide max old space size via resourceLimits option. https://nodejs.org/dist/latest-v12.x/docs/api/worker_threads.html#worker_threads_worker_resourcelimits

Issue #334, opened 2018-01-22
Cannot find module '--max-old-space-size=...'


BenGermany:
Hi there,

after some month of abstinence from programming i am working on my last project and wondering while using pkg to build a binary exe on windows 10 (pkg -t nodeX-win filename.js). I tested node 8, 9, x64, all, - and i found none that runs.

Some month ago, the same code works (i am irritated), now, when i run the executable, i got the error "Cannot find module '--max-old-space-size=...'". At this moment, i just forked a new child with execArgv some memsize.

![image](https://user-images.githubusercontent.com/26675531/35243522-780afe82-ffbd-11e7-9c39-e2416c0bee18.png)

...module.js:536 throw err...

I tested with pkg locally, global and node (7?) 8, 9,... via nvm-windows.

What i changed is the project-location and some structure, but why "Cannot find module ... max-old...", when it runs normally via "node filename.js"?.

Any suggestions?

THX

Update after some testing:

when i run the next script without forking, there is the next error:

![image](https://user-images.githubusercontent.com/26675531/35289608-15a41818-0068-11e8-907a-714828f92131.png)

This definitively runs some month ago - whats wrong here?

I copied all relevant files in new folders, running npm install and the files itself - with node everything is correct, with pkg - error.

Nobody else?

PR #1157, opened 2021-04-30
support execArgv when forking with child_process

paul-marechal:
Implement a workaround to convert opts.execArgv into
opts.env.NODE_OPTIONS when calling child_process.fork.

Note that this will most likely require you to package apps using
--no-bytecode because some node options will prevent you from running
the packaged bytecode with new options.

Fixes https://github.com/vercel/pkg/issues/334
Fixes https://github.com/vercel/pkg/issues/909

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -1952,6 +1952,7 @@ function payloadFileSync(pointer) {
     execFileSync: childProcess.execFileSync,
     exec: childProcess.exec,
     execSync: childProcess.execSync,
+    fork: childProcess.fork,
   };
 
   function setOptsEnv(args) {
@@ -1963,6 +1964,16 @@ function payloadFileSync(pointer) {
     }
     const opts = args[pos];
     if (!opts.env) opts.env = _extend({}, process.env);
+    if (Array.isArray(opts.execArgv)) {
+      // fork: The patched Node.js binary prevents us from passing execArgv on
+      // the cli. Instead we'll pass those options through the environment.
+      // https://nodejs.org/docs/latest/api/cli.html#cli_node_options_options
+      opts.env.NODE_OPTIONS =
+        typeof opts.env.NODE_OPTIONS === 'string'
+          ? `${opts.env.NODE_OPTIONS} ${opts.execArgv.join(' ')}`
+          : opts.execArgv.join(' ');
+      delete opts.execArgv;
+    }
     if (opts.env.PKG_EXECPATH === 'PKG_INVOKE_NODEJS') return;
     opts.env.PKG_EXECPATH = EXECPATH;
   }
@@ -2069,6 +2080,12 @@ function payloadFileSync(pointer) {
     setOptsEnv(args);
     modifyLong(args, 0);
     return ancestor.execSync.apply(childProcess, args);
+  };
+
+  childProcess.fork = function fork() {
+    var args = cloneArgs(arguments);
+    setOptsEnv(args);
+    return ancestor.fork.apply(childProcess, args);
   };
 })();
 
--- a/test/test-1157-fork-exec-argv/main.js
+++ b/test/test-1157-fork-exec-argv/main.js
@@ -1,0 +1,36 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const assert = require('assert');
+const utils = require('../utils.js');
+
+assert(!module.parent);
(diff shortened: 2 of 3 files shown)


marechal-p:
There might be a smarter way to go about this by modifying `pkg-fetch` Node patches, but it's a bit much for me to ramp up on... In the meantime, this workaround unblocked me enough to be able to package an application that would potentially rely on execArgv when forking.

erossignon (changes requested):
a unit test would be needed to demonstrate the effect and improved behavior, and prevent future regression

marechal-p:
@erossignon you are right, done!

prelude/bootstrap.js
```diff
@@ -1630,6 +1631,13 @@ function payloadFileSync(pointer) {
     const opts = args[pos];
     if (!opts.env) opts.env = require('util')._extend({}, process.env);
     if (opts.env.PKG_EXECPATH === 'PKG_INVOKE_NODEJS') return;
+    if (Array.isArray(opts.execArgv)) {
+      // fork: The patched Node.js binary prevents us from passing execArgv on
+      // the cli. Instead we'll pass those options through the environment.
+      // https://nodejs.org/docs/latest/api/cli.html#cli_node_options_options
+      opts.env.NODE_OPTIONS = opts.execArgv.join(' ');
+      delete opts.execArgv;
+    }
```
marechal-p, line 1640: First, I know this is a workaround so I don't know how welcome that kind of change is.

Second, would it make sense to move that logic right above the `PKG_INVOKE_NODEJS` check?

marechal-p, line 1640: First, I know this is a workaround so I don't know how welcome that kind of change is.

Second, would it make sense to move that logic right above the `PKG_INVOKE_NODEJS` check?

marechal-p:
I noticed the tests were failing because I wrongfully guessed Node's default memory limits. I now use `--disallow-code-generation-from-strings` to check if Node options are correctly passed and used by the child runtime.

marechal-p:
@robertsLando thanks for merging master into this PR. I rebased on master instead and force pushed.

I don't understand why the tests are behaving differently on the CI compared to my machine?

marechal-p:
I removed the bit that failed on CI, it was checking the output but it's not really necessary as long as we already detect that something wrong happened by checking for exit code 4.

Hopefully it works on the CI now...

marechal-p:
CI was green, I just rebased on master.

robertsLando:
Let's see tests result

hacke2:
When can merged, thank you

paul-marechal:
I rebased on master and fixed merge conflicts.

paul-marechal:
@erossignon @robertsLando bump

robertsLando:
Let's see hoe test goes first

SBell6hf:
This PR also fixes Issue https://github.com/vercel/pkg/issues/1438. I would be grateful if it's merged.

lib/packer.ts
```diff
@@ -47,7 +47,7 @@ function hasAnyStore(record: FileRecord) {
 interface PackerOptions {
   records: FileRecords;
   entrypoint: string;
-  bytecode: string;
+  bytecode: boolean;
```
robertsLando, line 50: Why this change?

robertsLando (changes requested):

git9am:
This PR looks helpful. Any chance to merge it?

paul-marechal:
> Why this change?

@robertsLando it was a left-over of when I was trying to understand the code to get this PR to work.

lib/packer.ts
paul-marechal: It was a left-over of when I was trying to understand the code to get this PR to work.

git9am:
Similar issues exist while calling `child_process.spawn` to start another node process with arguments.

Minimal code snippets to reproduce

```js
const { spawn } = require('child_process');

const node = spawn('node', ['--version']);
node.stdout.on('data', data => console.log(`[node] stdout: ${data}`));
node.stderr.on('data', data => console.log(`[node] stderr: ${data}`));
```

paul-marechal:
@git9am I guess this is because `bootstrap.js` patches `spawn` to interpret `node` as the current packaged binary.

Though I have no clue how to handle this, because there's no easy way to know what's an execArgv and what's not from the argument list passed :/

PR #1158, opened 2021-05-01
fs.createReadStream and --compress

erossignon:
@Hypfer , this should fix issue found with compressed pkg file and streaming file from snapshot.


prelude/bootstrap.js
```diff
@@ -565,10 +569,53 @@ function payloadFileSync(pointer) {
   // open //////////////////////////////////////////////////////////
   // ///////////////////////////////////////////////////////////////
 
+  function removeTemporaryFolderAndContent(folder) {
+    if (NODE_VERSION_MAJOR <= 14) {
+      if (NODE_VERSION_MAJOR <= 10) {
+        // folder must be empty
+        for (const f of fs.readdirSync(folder)) {
+          fs.unlinkSync(path.join(folder, f));
```
robertsLando, line 577: Could it be that `f` is a directory? If so this would fail

prelude/bootstrap.js
```diff
@@ -565,10 +569,53 @@ function payloadFileSync(pointer) {
...
+      }
+    } else {
+      fs.rmSync(folder, { recursive: true });
+    }
+  }
+  const temporaryFiles = {};
+  const os = require('os');
+  const tmpFolder = fs.mkdtempSync(path.join(os.tmpdir(), 'pkg-'));
+  process.on('beforeExit', () => {
+    removeTemporaryFolderAndContent(tmpFolder);
```
robertsLando, line 591: Is this really needed? Couldn't we keep files in temporary folder for quicker startup?

robertsLando (changes requested):

prelude/bootstrap.js
erossignon: We are controlling what goes in this temp folder, there are only flat files by design.

prelude/bootstrap.js
erossignon: -This is a boy scout attitude. Leave the place tidier than you found it.
- Those files potentially are exposed in clear, 
- Arguably, those files could be corrupted later on somehow, the single source of truth is in the packed binary.
- Only files that are  randomly accessed are copied uncompressed in this folder. I made sure that file that are read globally (readFileSync) or streamed (createReadStream) are processed in memory.  I would be inclined to leave it this way unless users complain about potential performance issue.
- having a persistent cache would require that the temp subfolder name is stable and unique, based on the executable signature (shall we calculated the md5/sha1 of the entire pkg.exe ?)

robertsLando: > having a persistent cache would require that the temp subfolder name is stable and unique, based on the executable signature (shall we calculated the md5/sha1 of the entire pkg.exe ?)

Could be an idea but yeah those files could be someway corrupted. I would like to also know @jesec and @leerob opinion about this

Hypfer: It might also be an option to generate a build-id uuid on build and bake that into the binary.

robertsLando (approved):

prelude/bootstrap.js
```diff
@@ -585,13 +632,37 @@ function payloadFileSync(pointer) {
     }
   }
 
+  let bypassCompressCheckWhenCallbyCreateReadStream = false;
```
jesec, line 635: Is it possible to have a race condition with this "global" variable? 

What will happen if multiple threads call related functions, and some of them call `createReadStream` while others don't?

erossignon: I though of this as well. So far, my investigation shows that;
 - this javascript code is only used by the nodejs thread, meaning that there will be only one call of  createReadStream at a time.
 -  the fs.createReadStream  is calling fs.open without async nor callback call in between. this means that the function execution is atomic and that a call to fs.createReadStream leads to a call to fs.open without going back to the eventloop.
 
- obviously, this assumption could be defeated if the nodejs guys decide to change their implementations.

robertsLando: This may happen when using worker threads or clusters I think?

jesec:
@erossignon Please rebase the PR. Apparently, I am not able to do this for you: 

```
remote: Permission to node-opcua/pkg.git denied to jesec.
```

Despite some concerns about race conditions, I think we can merge this PR, as the change appears to be self-contained, and should not hurt `no compress` case.

prelude/bootstrap.js
erossignon: I think that each worker thread will have their own copy of the global variable.

robertsLando:
Could this be merged?

erossignon:
Sure !

PR #1173, opened 2021-05-10
Add support for fs.realpath.native

Infern1:
When using Node.js fs-extra realpath.native is needed

See issue [here](https://github.com/jprichardson/node-fs-extra/issues/901)


robertsLando (approved):
LGTM

PR #1197, opened 2021-05-27
Improve README: remove dead badges, update the target list, add notes about building for different archs

jesec:


robertsLando (approved):
LGTM

jesec:
Additionally, we should have a prebuilt binary compatibility table somewhere. 

I think I will put that to the pkg-fetch repo.

leerob:
Thank you!

Issue #1192, opened 2021-05-24
5.2.x breaks express.js route handling logic?


nlfiedler:
The 5.1.0 version of pkg works fine for compiling and running our application, but the 5.2.1 (and maybe 5.2.0) changed in a way that the route handling in Express.js fails with a strange error.

```
::ffff:127.0.0.1 - - [24/May/2021:21:49:04 +0000] "GET / HTTP/1.1" 500 148
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type number (0)
    at validateString (internal/validators.js:124:11)
    at Object.resolve (path.js:1039:7)
    at Object.process.pkg.path.resolve (pkg/prelude/bootstrap.js:477:25)
    at View.lookup (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/view.js:114:32)
    at new View (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/view.js:94:20)
    at Function.render (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/application.js:570:12)
    at ServerResponse.render (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/response.js:1012:7)
    at /snapshot/helix-auth-svc-2021.1.0/lib/app.js:116:7
    at Layer.handle_error (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/router/layer.js:71:5)
    at trim_prefix (/snapshot/helix-auth-svc-2021.1.0/node_modules/express/lib/router/index.js:315:13)
```

The code in question is at https://github.com/perforce/helix-authentication-service although I doubt any of that is relevant.

Let me know what helpful information I can provide. I realize that what has been said so far is not much to go on.

robertsLando:
@jesec ?

jesec:
The only significant change in `5.2.x` that may be related to this issue is the compression. 

Other changes are just routine updates (bump Node.js versions, etc.) or macOS-only (codesigning, M1 support, etc.), which are quite unlikely to be related. #1173 is indeed a global functional change, but I don't think that's related. 

cc: @erossignon

jesec:
@nlfiedler I still need a minimum repo to reproduce. I can't reproduce it in my own Express.js project.

erossignon:
[...]

Issue #1191, opened 2021-05-24
Error thrown with pkg@5.2.1 Error: EISDIR


cowlicks:
This error started being thrown this morning. And only with 5.2.1. It is not thrown on 5.2.0.
```
pkg/prelude/bootstrap.js:1614
      throw error;
      ^

Error: EISDIR: illegal operation on a directory, read
    at error_EISDIR (pkg/prelude/bootstrap.js:553:17)
    at readFileFromSnapshot (pkg/prelude/bootstrap.js:894:33)
    at Object.readFileSync (pkg/prelude/bootstrap.js:938:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1095:18)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at Module.require (pkg/prelude/bootstrap.js:1593:31)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/snapshot/app/node_modules/sqlite3/lib/sqlite3-binding.js:4:15) {
  errno: -21,
  code: 'EISDIR',
  path: '/snapshot/app/node_modules/sqlite3/lib///',
  pkg: true
}
there was a problem running the executable
```

tinoks:
Same error when updating to 5.2.0 from 5.1.0.


```
pkg/prelude/bootstrap.js:1614
      ^

Error: EISDIR: illegal operation on a directory, read
    at error_EISDIR (pkg/prelude/bootstrap.js:553:17)
    at readFileFromSnapshot (pkg/prelude/bootstrap.js:894:33)
    at Object.readFileSync (pkg/prelude/bootstrap.js:938:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1095:18)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at Module.require (pkg/prelude/bootstrap.js:1593:31)
    at require (internal/modules/cjs/helpers.js:88:18)
    at bindings (D:\snapshot\ogcapi-fast2\node_modules\bindings\bindings.js:112:48) {
  code: 'EISDIR',
[...]

PR #1199, opened 2021-05-31
Squashed revert "allow to compress files in virtual file system (#1115)"

jesec:
This reverts commit 82ce62558e55360f1a7815e5899f5e88a1ce7ddb.
This reverts commit ea231969d6dedb31a91dbd2278e1387aea3d77d2.
This reverts commit ea7d72bc2052e717bd89ab1714d5894dd5e9bfa7.

Apparently the compression feature needs more work. 

I can confirm that this revert at least fixes `const db = require('better-sqlite3')("test.sqlite");`, mentioned in https://github.com/vercel/pkg/issues/1191#issuecomment-848511522. 

Bug: #1191, #1192


leerob (approved):

robertsLando (approved):

jesec:
@leerob We can release `5.2.2`.

erossignon:
That's harsh !  Shall we fix root cause of #1191, #1192 instead ?

jesec:
> That's harsh ! Shall we fix root cause of #1191, #1192 instead ?

I understand. However, the bugs are there for a week. Since none of us dealt with the issue in time, the most logical and straightforward option is to fix a severe bug by reverting an optional feature.

PR #1200, opened 2021-05-31
Compression with fix for  #1191 & #1192 

erossignon:


erossignon:
fixes also #1202

jesec:
I am doing a line-by-line review. It would take some time. 

The change is giant. I think it would make it easier to review if we can split some less related changes to separate commits, and possibly get them merged first. 

Additionally, I am thinking about a fd-mapping approach to replace the `uncompressExternallyAndOpen`.

erossignon:
The pr comes in 3 commit
A big one that reverts the revert of the 3 commits of compression that was already validated and 2 smaller commits that fixes #1191 and #1192.
I suggest to look at each individual commits .

jesec:
I would like to re-review the compression change. A giant change is not reviewable, and it would be easier to review if we split it into several smaller changes.

We can also get some changes that are not directly related to compression merged first.

erossignon:
That make sense to unsquash and reorganise the commit while we are here! Agreed

jesec:
> That makes sense to unsquash and reorganise the commit while we are here! Agreed

Great! Please review #1204 when you have time.

drothmaler:
Nice work, I'm really looking forward to the compression feature. I'm using Globalize.js and it requires a lot of CLDR JSON data, that increases the size of the bundle a lot, so compression would come in very handy :-) 

I think it would be good to also be able to configure the compression using package.json / x.config.json, like you can do for all the other parameters...

Hypfer:
Unfortunately, the code as it is here now doesn't seem to work with my application.
When I try to request one of the static files served by expressjs, I get one of these messages on stdout

```
Decompression failed
    at BrotliDecoder.zlibOnError [as onerror] (node:zlib:190:17) {
  errno: -2,
  code: 'ERR_RESERVED',
  expose: false,
  statusCode: 500,
  status: 500
}
```
```
incorrect header check
    at Zlib.zlibOnError [as onerror] (node:zlib:190:17) {
  errno: -3,
  code: 'Z_DATA_ERROR',
  expose: false,
  statusCode: 500,
  status: 500
}
```

I'd provide more logs but there are none

erossignon:
@Hypfer  can you provide a minimum gist that reproduce the issue ?

Hypfer:
@erossignon Sure

https://github.com/Hypfer/pkg-compression-test-example

erossignon:
> @erossignon Sure
> 
> https://github.com/Hypfer/pkg-compression-test-example

@Hypfer Thank you, 
I found out that if you remove --targets node16, it works fine ....

It must be some change in behavior in node16 for the compress lib

jesec:
@erossignon :

I am not able to update the PR directly. Please pull the rebased commit from https://github.com/vercel/pkg/tree/pr/compression.

jesec:
I noticed that you added a function `findVirtualFileSystemEntry`. However, later lines no longer use normalized `path_`. I believe that you may be relying on the incoming-parameter-modifying behavior of certain functions (needlessly to say, is an anti-pattern). Is that so, or is this change intended? 

image

erossignon:
> findVirtualFileSystemEntry

Yes, I am aware of this. Before VHS fileame compression, the two form was equivalent( 
At the end, this is to create a error message that referees to the filename the used want to deal with, It is better to use the human readable path_ rather that path_vfs that could be cryptic if compress option is used.

erossignon:
@Hypfer  the lastest version of this PR  has fixed the issue you found.

Hypfer:
I can confirm that the issue discovered has been fixed 👍

Unfortunately though, I did discover two new issues. I've updated [the example repo](https://github.com/Hypfer/pkg-compression-test-example) to feature these.

<details>
<summary>Click me for details</summary>
This is what it should look like:

image

This is how it actually looks like:
image

And here's the console output:
image
</details>

jesec:
> > findVirtualFileSystemEntry
> 
> Yes, I am aware of this. Before VHS filename compression, the two forms were equivalent
> In the end, this is to create an error message that refers to the filename, It is better to use the human-readable `path_` rather than `path_vfs` that could be cryptic if the compress option is used.

Got it. However, the function is modifying the incoming parameter. So the "easier to debug" part would not work at all. 

Additionally, although in this case, the path is merely used for error message, there are other cases where the path has functional implications. 

Can you take a look at those instances? We should definitely avoid relying on incoming-parameter-modifying behavior. I also plan to remove all such cases and enforce the `no-param-reassign` rule.

erossignon:
> Got it. However, the function is modifying the incoming parameter. So the "easier to debug" part would not work at all.
OK. I fixed in a separate commit 'do not mess with function paramters' to facilitate the review, We'll squash if needed.

erossignon:
> Unfortunately though, I did discover two new issues. I've updated [the example repo](https://github.com/Hypfer/pkg-compression-test-example) to feature these.

@Hypfer  => fixed now

ayosec:
I tried this patch with [mathjax-node-cli](https://github.com/mathjax/mathjax-node-cli) and I didn't find any issue.

Something unexpected is how slow is Brotli:

| Compression | Time to build | Final size |
|-------------|--------------:|-----------:|
| None        | 8 seconds     | 111 Mib    |
| Gzip        | 15 seconds    | 62 Mib     |
| Brotli      | 142 seconds   | 58 Mib     |


<details>
<summary>Steps to reproduce</summary>

1. Download and build this patch.

    ```console
    $ docker run --rm -ti node bash

    # git clone --depth 1 https://github.com/vercel/pkg.git /tmp/pkg

    # cd /tmp/pkg

    # git fetch origin refs/pull/1200/head:pr1200

    # git checkout pr1200

    # git log -1 --pretty=%H
    31c95f8eb86599f06d786c47cb47d7aef5a709c5

    # yarn install && yarn build
    ...  
    Done in 3.79s.
    ```

2. Install `mathjax-node-cli` and build it against the compression methods.

    ```console
    # npm install -g mathjax-node-cli

    # SRC=/usr/local/lib/node_modules/mathjax-node-cli/bin/tex2svg

    # time node lib-es5/bin.js -t node16-linux-x64 -o bin $SRC
    > pkg@5.2.1

    real        0m8.245s
    user        0m10.345s
    sys         0m1.404s

    # time node lib-es5/bin.js -C Gzip -t node16-linux-x64 -o bin-gz $SRC
    > pkg@5.2.1
    compression:  GZip

    real        0m14.755s
    user        0m17.924s
    sys         0m1.698s

    # time node lib-es5/bin.js -C Brotli -t node16-linux-x64 -o bin-br $SRC
    > pkg@5.2.1
    compression:  Brotli

    real        2m22.321s
    user        2m25.101s
    sys         0m2.064s

    # ls -1sh bin*
    111M bin
     58M bin-br
     62M bin-gz
    ```

3. Test the generated binaries.

    ```console
    # time ./bin '\frac{2}{x}' > 1

    real        0m0.707s
    user        0m1.225s
    sys         0m0.041s

    # time ./bin-gz '\frac{2}{x}' > 2

    real        0m0.852s
    user        0m1.355s
    sys         0m0.079s

    # time ./bin-br '\frac{2}{x}' > 3

    real        0m0.881s
    user        0m1.387s
    sys         0m0.056s

    # cmp 1 2 && cmp 2 3 && echo OK
    OK

    # head -c 50 1
    <svg xmlns:xlink="http://www.w3.org/1999/xlink" wi
    ```

</details>

---

Also, please consider adding [LZ4](https://en.wikipedia.org/wiki/LZ4_(compression_algorithm)). Since most of the compressed data are JavaScript files, with maximum compression level it is close to Gzip, but much faster.

Example:

```console
# find -type f -name '*js' -exec cat {} + > /tmp/big.js

# ls -sh /tmp/big.js
90M /tmp/big.js

# time lz4 -9 < /tmp/big.js | wc -c | numfmt --to=iec
20M

real	0m2.814s
user	0m2.787s
sys	0m0.026s

# time gzip -9 < /tmp/big.js | wc -c | numfmt --to=iec
17M

real	0m5.421s
user	0m5.380s
sys	0m0.047s
```

erossignon:
@ayosec  LZ4  is not built-in nodejs and would be difficult to implement. Feel free to propose a PR if you are inspired.

Hypfer:
@erossignon Can confirm. Thank you!

Q: Since pkg is now using os.tmpdir() which often is a ramdisk to uncompress all/some(?) files, is there any rough estimate regarding memory overhead of using compression?

erossignon:
> @erossignon Can confirm. Thank you!
> 
> Q: Since pkg is now using os.tmpdir() which often is a ramdisk to uncompress all/some(?) files, is there any rough estimate regarding memory overhead of using compression?


- a/ files that are accessed with `fs.readFile` or `fs.readFileSync` are uncompressed on the fly, in memory. This is the case for all embedded javascript script files. 
- b/ files that are accessed with `fs.open/fs.openSync` & `fs.read/fs.readSync` needs to be deflated in a temporary folder beforehand as they can potentially read by random access. 
- c/ files that are accessed with `fs.createReadStream` used to be uncompressed on the fly, in memory in my first attempt, At this time ,  `fs.createReadStream` fallback to the `fs.openSync` & `fs.read` scenario and deflated in the tempfolder.

- d/ binary modules are always copied to a physical disc (regardless of compress flag)

-  scenario b/ & c/ are likely used when the application tries to access 'assets' files.


Note on c:
In my first attempt, createReadStream did uncompress on the fly, in memory. but the mechanism was brittle and convoluted.
On top of this, it did work when createReadStream was passed a non-zero start position and some variation of the internal implementation exists between versions of the node.   I have another idea to improve this but I'll do this in a different PR as this one is quite big now.

- if your packed application doesn't contain asset files, nor binary modules then the temp folder is likely not to be used.

test/test-12-compression/.gitignore
```diff
@@ -0,0 +1 @@
+ouput-*
```
jesec, line 1: typo

test/test.js
```diff
@@ -42,8 +42,11 @@ function joinAndForward(d) {
 
 const list = [];
 
+console.log('FLAVOR = ', flavor);
 if (flavor === 'only-npm') {
   list.push(joinAndForward('test-79-npm/main.js'));
+  list.push(joinAndForward('test-1191/main.js'));
```
jesec, line 48: IMO it is cleaner to move these two tests to `test-79-npm` folder, so we can remove the special logics here (and several lines below).

Maybe `better-sqlite3` and `express-route` folders?

test/test-79-npm/homebridge/homebridge.js
```diff
@@ -4,7 +4,6 @@ process.argv.push('--help');
 require('homebridge/lib/cli.js'); // dont run. only load
 var Server = require('homebridge/lib/server.js').Server;
 var server = new Server();
-
```
jesec, line 7: unnecessary and unrelated change.

lib/packer.ts
```diff
@@ -1,8 +1,8 @@
 /* eslint-disable complexity */
 
 import assert from 'assert';
-import fs from 'fs-extra';
-import path from 'path';
+import * as fs from 'fs-extra';
+import * as path from 'path';
```
jesec, line 5: This does not fit with other uses of `fs-extra` and `path`. 

and unnecessary?

lib/index.ts
```diff
@@ -625,12 +652,14 @@ export async function exec(argv2: string[]) {
       await mkdirp(path.dirname(target.output));
     }
 
+    const slash = target.platform === 'win' ? '\\' : '/';
     await producer({
       backpack,
       bakes,
-      slash: target.platform === 'win' ? '\\' : '/',
+      slash,
```
jesec, line 659: unnecessary?

lib/index.ts
```diff
@@ -274,7 +277,31 @@ export async function exec(argv2: string[]) {
 
   const forceBuild = argv.b || argv.build;
 
-  // _
```
jesec, line 277: The section comment `// _` is probably replaced erroneously. this should be moved to line 305 instead.

lib/compress_type.ts
```diff
@@ -0,0 +1,6 @@
+
```
jesec, line 1: EOF and white line.

prelude/bootstrap.js
```diff
@@ -1489,7 +1802,7 @@ function payloadFileSync(pointer) {
 
     if (entityBlob) {
       const options = {
-        filename: path_vfs,
+        filename: filename_,
```
jesec, line 1805: Are you sure that we don't need to use the normalized path here?

prelude/bootstrap.js
```diff
@@ -24,6 +25,67 @@ const path = require('path');
 const { promisify, _extend } = require('util');
 const { Script } = require('vm');
 const { tmpdir } = require('os');
+const util = require('util');
+const {
+  gunzip,
```
jesec, line 30: keep it sorted.

prelude/.prettierrc.yml
```diff
@@ -0,0 +1,2 @@
+printWidth: 80
```
jesec, line 1: These overrides are not really necessary?

prelude/bootstrap.js
```diff
@@ -24,6 +25,67 @@ const path = require('path');
...
+  gunzipSync,
+  brotliDecompressSync,
+  brotliDecompress,
+} = require('zlib');
+
+/**
+ * note: developers may activate the doDebug flag to enable a trace in the console
```
jesec, line 37: I think the debug routines should be a separate PR.

prelude/bootstrap.js
```diff
@@ -236,11 +343,10 @@ function projectToFilesystem(f) {
 function projectToNearby(f) {
   return path.join(xpdn, path.basename(f));
 }
-
```
jesec, line 239: IMO it is better to avoid line change in unrelated functions.

.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  // note you must disable the base rule as it can report incorrect errors
```
jesec, line 2: These changes are no longer needed after #1211.

I tested the PR without changes to .eslintrc, and `yarn lint` reports no error.

jesec: I mean you can simply not change the file at all.

test/test.js
erossignon: I've added a specific test for better-slite in the test-79 series

however, I kept test test-1191  & test-1992  as they are testing different things.
I simply enabled them to run when platform === "host" as the incorporated binary module would be mismatching the targeted node version.

.eslintrc
```diff
@@ -9,8 +9,10 @@
     "consistent-return": "off",
     "no-restricted-syntax": "off",
     "import/prefer-default-export": "off",
-    "camelcase": "off"
-  },
+    "camelcase": "off",
+    "no-shadow": "off",
+    "@typescript-eslint/no-shadow": ["error"]
+    },
```
jesec, line 15: These changes are not needed. You can simply `git checkout master .eslintrc` to reset the file to the original.

jesec (approved):

jesec:
@leerob We can release v5.2.2.

Bunos07 (approved):
Help 🥺 I didn't know kung Anu ginagawa ko gusto kolang matutuoo

PR #1211, opened 2021-06-04
Enforces more ESLint rules

jesec:


erossignon (commented):
I am not sure that replacing (for ...of ) loop with forEach is useful.
My experience shows that It could be slightly slower and sometimes get in your way during debugging as you cannot easily step in the lambda code. Also some pitfalls and traps during refactoring => continue=>return etc ...

https://thecodebarbarian.com/for-vs-for-each-vs-for-in-vs-for-of-in-javascript

jesec:
> I am not sure that replacing (for ...of ) loop with forEach is useful.
> My experience shows that It could be slightly slower and sometimes get in your way during debugging as you cannot easily step in the lambda code. Also some pitfalls and traps during refactoring => continue=>return etc ... introducing hard to detect bugs
> 
> https://thecodebarbarian.com/for-vs-for-each-vs-for-in-vs-for-of-in-javascript

I see. I will drop it for now.

robertsLando (approved):

Issue #1207, opened 2021-06-03
DEBUG_PKG does not work when project is not on drive C: on Windows


FlorentMasson:
I get this error after I follow the instructions [here](https://github.com/vercel/pkg#advanced)

```
D:\project\server_exe>server-win.exe
------------------------------- virtual file system
C:\snapshot
pkg/prelude/bootstrap.js:322
  if (error) throw error;
             ^

Error: Directory 'C:\**\' was not included into executable at compilation stage. Please recompile adding it as asset or script.
    at error_ENOENT (pkg/prelude/bootstrap.js:539:17)
    at readdirFromSnapshot (pkg/prelude/bootstrap.js:1057:29)
    at Object.readdirSync (pkg/prelude/bootstrap.js:1083:19)
    at dumpLevel (pkg/prelude/bootstrap.js:2066:18)
    at installDiagnostic (pkg/prelude/bootstrap.js:2111:23)
    at pkg/prelude/bootstrap.js:2142:3
    at pkg/prelude/bootstrap.js:2144:3
    at readPrelude (internal/bootstrap/pkg.js:31:12)
    at internal/bootstrap/pkg.js:36:18
    at internal/bootstrap/pkg.js:43:4 {
  errno: -4058,
  code: 'ENOENT',
  path: 'C:\\snapshot',
  pkg: true
}
```

It works fine when I copy the same project to C:

erossignon:
@FlorentMasson  can you provide a minimalist test that reproduces the behavior ? 
I try compiling a simple test script and move it to a different drive but I cannot reproduce it.

FlorentMasson:
Is this enough as a minimalist test?
To clarify, I'm compiling from a different drive (D:), not on C: then moving the executable

```
D:\>pkg -v
5.2.1

D:\>md testpkg

D:\>cd testpkg

D:\testpkg>echo console.log('hello'); > test.js

D:\testpkg>pkg -d test.js
> pkg@5.2.1
> Targets not specified. Assuming:
  node12-linux-x64, node12-macos-x64, node12-win-x64
> [debug] Bytecode of D:\testpkg\test.js is added to queue.
> [debug] Stat info of D:\testpkg\test.js is added to queue.
> [debug] Directory D:\testpkg is added to queue.
> [debug] Stat info of D:\testpkg is added to queue.
> [debug] Directory D:\ is added to queue.
> [debug] Stat info of D:\ is added to queue.
> [debug] The file was included as bytecode (no sources)
[...]

PR #1212, opened 2021-06-06
fix issue #1207, Windows: pkg run from disc other than C:

erossignon:


lib/common.ts
```diff
@@ -119,8 +119,9 @@ function injectSnapshot(file: string) {
       // C:\
       file = file.slice(0, -1);
     }
-
-    return `${file[0]}:\\snapshot${file.slice(2)}`;
+    // by convention, on windows we use C:\\snapshot
+    const pathInSnapshot = `C:\\snapshot${file.slice(2)}`;
+    return pathInSnapshot;
```
jesec, line 124: return directly here?

lg2m in principle.

jesec (approved):

lib/common.ts
erossignon: Yes, , I had a console.log at some point while developing, hence I store the intermediate result in a variable. 

Btw, I am struggling to find a way to create a non-regression test for this issue. as I am not sure we can easily create an alternative drive on the CI/CD Windows build the image.

jesec: You can mount a VHD.

erossignon: I found that `subst H: C:\MyFolder` will work.

lib/common.ts
```diff
@@ -119,8 +119,8 @@ function injectSnapshot(file: string) {
       // C:\
       file = file.slice(0, -1);
     }
-
-    return `${file[0]}:\\snapshot${file.slice(2)}`;
+    // by convention, on windows we use C:\\snapshot
+     return  `C:\\snapshot${file.slice(2)}`;
```
jesec, line 123: extraneous whitespace here.

PR #1214, opened 2021-06-07
add debugging capabilities in bootstrap.js

erossignon:


prelude/bootstrap.js
```diff
@@ -25,6 +25,60 @@ const { promisify, _extend } = require('util');
...
+        lastArgument.apply(null, arguments);
+      };
+      f.apply(this, arguments);
+    } else {
+      const ret = f.apply(this, arguments);
+      debugIndent -= 1;
+      return ret;
+    }
+  };
+}
```
robertsLando, line 67: Instead of a debugWrap function that could mess up the function calling context I would suggest to create a function that receives the function context as parameter.

```
function debugCall(context, args) {...}

// inside functions:
debugCall(this, arguments)
```

erossignon: Can you be more specific?

robertsLando: Now it is:

debugWrap(function f() {})

I suggest:

function f() {
 debugCall(this, arguments)
}

erossignon: OK , I see.

I don't think that this would work. 

We need to capture the name of the wrapped function.

Also we want to have no effect on the code ( same behavior as before ) when debug === false , in which case debugWrap is a passthrough

robertsLando: Ok was just an idea :)

PR #1216, opened 2021-06-07
add Cancel Workflow action

erossignon:


jesec:
I don't think there is a need to roll CI. Just let them run.

erossignon:
I kept this PR in draft mode as it was an experiment.
This is supposed to kill the previous job on the same branch only.
I figure out that it doesn't work for a build that takes place from a forked repo as the user needs to have some access right on the main repo.
So, I'm inclined not to waste too much time on this for now.
@jesec , let's close.

PR #1246, opened 2021-07-04
Add docs for automatic releases w. GitHub Actions

karlhorky:
Hi there, thanks for `pkg`, really amazing!

Wanted to test the waters on documenting usage of GitHub actions to automate building using `pkg` and uploading the executables as assets to a GitHub release.

Example here:

- Workflow: https://github.com/jonaschlegel/lgo-agisoft-converter/blob/main/.github/workflows/release.yml
- Release: https://github.com/jonaschlegel/lgo-agisoft-converter/releases/tag/v1.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,65 @@ await exec(['app.js', '--target', 'host', '--output', 'app.exe']);
 // do something with app.exe, run, test, upload, deploy, etc
 ```
 
+## Automate Releases with GitHub Actions
+
+[GitHub Actions](https://docs.github.com/en/actions) can be used to automatically run `pkg` when a specific event has occurred.
+
+For example, to automatically run `pkg` to create executables and upload them as assets to a new release on GitHub every time a new Git tag is pushed, try setting up [`softprops/action-gh-release`](https://github.com/softprops/action-gh-release) as follows:
+
+1. Add a new file in your project called `.github/workflows/release.yml` (create the directories `.github/workflow` in case they don't exist in your project) and add this file content:
+
+```yaml
+name: Build and release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+      - run: yarn --frozen-lockfile
+      - run: npm install --global pkg
+      - name: Build
+        run: pkg index.js --output your-program-name-here --targets linux,macos,win
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            your-program-name-here-linux
+            your-program-name-here-macos
+            your-program-name-here-win.exe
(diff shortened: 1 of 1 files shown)


jesec:
Thanks for the contribution. It is a very good tutorial.

Unfortunately, it would be inappropriate for us to endorse a specific CI service.

Additionally, naturally people assume that project maintainers provide support to contents in README of the project. However, we do not have the capacity to support Github Actions, and we will not be able to answer everyone's (in our view, out-of-scope) "how to automate my workflow" questions. 

In fact, there are many other ways to integrate `pkg` with the workflow of the project. A way that does not tie to a specific CI provider is to simply add a "build-pkg` command to `package.json`, so anyone (incl. any CI provider one may choose) can run `npm run build-pkg` to build the pkg executables.

karlhorky:
Ah ok, would it be accepted if there were examples for other CI platforms? Also with a disclaimer that there is no support provided for this?

The value that I see is in the ability of copy + pasting this without having to research the different parts and how to plug them together (which can take hours)

jesec:
> Ah ok, would it be accepted if there were examples for other CI platforms? Also with a disclaimer that there is no support provided for this?
> 
> The value that I see is in the ability of copy + pasting this without having to research the different parts and how to plug them together (which can take hours)

That kinda makes things worse, and we are now expected to provide support to not only Github Actions but also others. 

Well. We all know that people don't read disclaimer. ;) 

If things are in README, we are expected to provide (some degree of) support. I do not believe that we should do that, and we don't have resources for that. "How to automate your workflow" is quite out-of-scope for this project.

karlhorky:
Ok, then how about an `examples` or `recipes` folder, à la Next.js? Those are very clearly maintained by community members.

jesec:
> Ok, then how about an `examples` or `recipes` folder, à la Next.js? Those are very clearly maintained by community members.
> 
> Then the readme could link to this examples folder...

That would be a bit more acceptable. 

Keep in mind that Next.js is the main product and is maintained by a team of dedicated engineers. This project is not allocated the same level of resources and is mostly maintained in voluntary capacity. 

I believe that such out-of-scope tutorials/examples do not necessarily justify the maintenance burden, as a result.

karlhorky:
Ok, would such a PR be accepted? If not, then we can just leave it at that.

jesec:
> Ok, would such a PR be accepted? If not, then we can just leave it at that.

Unfortunately, no.

azerpas:
> Hi there, thanks for `pkg`, really amazing!
> 
> Wanted to test the waters on documenting usage of GitHub actions to automate building using `pkg` and uploading the executables as assets to a GitHub release.
> 
> Example here:
> 
> * Workflow: https://github.com/jonaschlegel/lgo-agisoft-converter/blob/main/.github/workflows/release.yml
> * Release: https://github.com/jonaschlegel/lgo-agisoft-converter/releases/tag/v1.0.0

Great example, thanks!

karlhorky:
@azerpas glad that it helped you!

Since it unfortunately won't be accepted here, I have published a separate repo with the guide: https://github.com/karlhorky/vercel-pkg-github-actions-release

PR #1249, opened 2021-07-07
Add classPrivateMethods plugin to babel parser

### LINKED ISSUE #1248: Error! This experimental syntax requires enabling the parser plugin: 'classPrivateMethods' (406:2)
I understand that a parser is being used at a low level here.
I've turned off bytecode generation and I really just want these files packed as-is as node can handle this syntax on the chosed target.

How do I turn this off?

hcldan:
Fixes #1248 

I'd like to be able to pass in my own version of babel parser, configured with my own set of plugins.


hcldan:
@leerob what's the story here for people wanting to pitch in? I never got any comments from a maintainer.

robertsLando:
@hcldan Sorry I may have missed this one

robertsLando (approved):

hcldan:
Thank you!

jesec:
I would recommend users to transform/transpile their projects first, if novel syntaxes or new JS features are used.

pkg supports older Node.js runtime so the built-in parser configuration is rather conservative. This makes sure that you would get an error if you attempt to use features that are only available in newer Node.js runtimes. Note that pkg only uses Babel to walk dependencies. It does not touch your source code.

`classPrivateMethods ` is only available on Node 12+. I can merge this after we dropped support for Node 8 and 10. Alternatively, we can have different parser configurations based on the Node target.

PR #837, opened 2020-01-30
Add support for Node native addons

geekuillaume:
This adds the supports for Nodejs native addons.

How it works:
- When detecting a `.node` file, it will package it like an asset and not output any warning.
- On run-time, when `require`ing a `.node` file, it will use the same code as classic files to detect if the file is in the snapshot or not. If it is, it will write it to a file in the `os.tmpdir()` directory.
- The filename of the temporary file will be passed to the rest of the chain.

Potential future optimizations:
- Write a patch for the `bindings` module to detect the correct modules.
- Delete the temporary files on process exit

Related #329 #619 #749 #663

PR #1273, opened 2021-08-03
Fix warnings when packaging with leveldown.

### LINKED ISSUE #1264: Warnings when bundling leveldown.
I'm trying to bundle a simple node program that includes some native modules. One of these modules is level witch includes leveldown.

When trying to use pkg to build I receive the following error.
```
pkg . --targets linux-x64 --no-bytecode --public-packages "*" --output linprog
> pkg@5.3.1
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/leveldown/prebuilds
  %2: path-to-executable/prebuilds
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/leveldown/prebuilds
  %2: path-to-executable/prebuilds
```
The built executable works if I include the prebuilds with it.

Now as far as I know, pkg should detect and include these native modules properly now. So I shouldn't need to include these manually. Is there any reason why this isn't working?

When I try to add the files as an asset in the package.json with:
```
"pkg": {
    "assets": [
      "node_modules/leveldown/**/*",
    ]
  }
```
I can confirm the files are being added when I check the log with the `--debug` flag.

The odd thing is, this project also includes utp-native witch seems to be bundled just fine.

tegefaulkes:
Fixes #1264.
Relates to #837

The goal of this PR is to remove warnings when packaging a project that includes the leveldown native module.
This change should;
* Remove warnings when packaging leveldown.
* packaged executable should run without needing to include the leveldown prebuilds directory with the executable.

### DIFF
--- a/dictionary/leveldown.js
+++ b/dictionary/leveldown.js
@@ -1,10 +1,3 @@
 'use strict';
 
-module.exports = {
-  pkg: {
-    patches: {
-      'binding.js': ['__dirname', "require('path').dirname(process.execPath)"],
-    },
-    deployFiles: [['prebuilds', 'prebuilds', 'directory']],
-  },
-};
+module.exports = {};


CMCDragonkai:
This is pretty quick fix. Can it be merged soon?

dsturnbull:
Resolvable in local config, e.g. package.json in the thing you're packaging up:

```
    "patches": {
      "node_modules/leveldown/binding.js": []
    }
```

This essentially removes the patch specified in `dictionary/leveldown.js` while we wait for this PR.

CMCDragonkai:
What is the `leveldown.js` dictionary used for anyway? Why is it there if packaging native leveldown results in a warning? Why not just remove that dictionary entirely for leveldown?

CMCDragonkai:
@robertsLando Changing to use `--no-dict leveldown` option did not work in our case. It resulted in: `> Error! Not more than one entry file/directory is expected`. Nothing else changed.

We're using something like:

```
        pkg . \
          --targets node${utils.nodeVersion}-linux-${arch} \
          --no-bytecode \
          --public \
          --public-packages "*" \
          --no-dict leveldown \
          --output out
```

CMCDragonkai:
It required `leveldown.js`.

Issue #1235, opened 2021-06-24
TypeError: Cannot destructure property `isRegExp` of 'undefined' or 'null'


gstepniewski:
Getting this error when running a package built with pkg 5.3.0:

```
pkg/prelude/bootstrap.js:23
const { isRegExp } = require('util').types;
        ^

TypeError: Cannot destructure property `isRegExp` of 'undefined' or 'null'.
    at __importDefault.default (pkg/prelude/bootstrap.js:23:37)
    at pkg/prelude/bootstrap.js:2110:3
    at readPrelude (bootstrap_node.js:148:16)
    at bootstrap_node.js:153:22
    at bootstrap_node.js:160:8
    at startup (bootstrap_node.js:161:6)
    at bootstrap_node.js:661:3
```

It was definitely working yesterday with pkg 5.2.1, not sure how to get more information to help with debugging this?

robertsLando:
@erossignon regression?

erossignon:
Strange , this destructuring is supposed to work on all supported nodejs version:

````
$nvm use 10
Now using node v10.21.0 (npm v6.14.7)
$node -e "const { isRegExp1 } = require('util').types; console.log(typeof isRegExp1);"
undefined
$nvm use 10
Now using node v10.21.0 (npm v6.14.7)
$node -e "const { isRegExp } = require('util').types; console.log(typeof isRegExp);"
function
$nvm use 12
Now using node v12.22.1 (npm v6.14.12)
$node -e "const { isRegExp } = require('util').types; console.log(typeof isRegExp);"
function
$nvm use 14
Now using node v14.17.0 (npm v6.14.13)
$node -e "const { isRegExp } = require('util').types; console.log(typeof isRegExp);"
function
$nvm use 16
Now using node v16.4.0 (npm v7.18.1)
$node -e "const { isRegExp } = require('util').types; console.log(typeof isRegExp);"
function
````

**Comment by erossignon** [comment]
@gstepniewski  what is the exact step to reproduce ( please provide the smallest possible example)

**Comment by gstepniewski** [comment]
Looks like a problem with node 8:

```
[~] node -v                                                                                                                                                                                                                                                          
v8.17.0
[~] node -e "const { isRegExp1 } = require('util').types; console.log(typeof isRegExp1);"                                                                                                                                                                            
[eval]:1
const { isRegExp1 } = require('util').types; console.log(typeof isRegExp1);
        ^

[...]

PR #1282, opened 2021-08-11
Update to support latest LTS node version(#1235)

paranlee:
This pull request solve #1235 error on pkg main example program.

This error occur because nodejs 8 has reached end-of-life end on the 31st 12/2019 is not recommended to be used in production anymore.

```
const { isRegExp1 } = require('util').types; console.log(typeof isRegExp1);
        ^

TypeError: Cannot destructure property `isRegExp1` of 'undefined' or 'null'.
    at [eval]:1:38
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at Object.runInThisContext (vm.js:139:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:653:30)
    at evalScript (bootstrap_node.js:479:27)
    at startup (bootstrap_node.js:180:9)
    at bootstrap_node.js:625:3
```

### DIFF
--- a/examples/express/package.json
+++ b/examples/express/package.json
@@ -3,14 +3,14 @@
   "license": "ISC",
   "bin": "index.js",
   "dependencies": {
-    "express": "4.15.2"
+    "express": "^4.17.1"
   },
   "pkg": {
     "assets": [
       "views/**/*"
     ],
     "targets": [
-      "node8"
+      "node16"
     ]
   }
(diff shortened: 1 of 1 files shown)


paranlee:
When pull request merged, then I would doing my #1285 issue. this is more generic and more newbie friendly!

PR #1321, opened 2021-09-29
fix: better support of `.node` files

### LINKED ISSUE #1075: Bundling sharp doesn't work anymore since updating to 4.5.0
Since updating to version 4.5.0 it is not possible to bundle sharp anymore. Previously this was fairly simple, by just copying the `node_modules/sharp` folder next to the build output. But this doesn't work anymore giving the following error:
```
Error:
Something went wrong installing the "sharp" module

File '/**/test/node_modules/sharp/build/Release/libvips-cpp.so.42' was not included into executable at compilation stage. Please recompile adding it as asset or script.

- Remove the "node_modules/sharp" directory then run
  "npm install --ignore-scripts=false --verbose" and look for errors
- Consult the installation documentation at https://sharp.pixelplumbing.com/install
- Search for this error at https://github.com/lovell/sharp/issues

    at Object.<anonymous> (/snapshot/test/node_modules/sharp/lib/constructor.js:34:9)
    at Module._compile (pkg/prelude/bootstrap.js:1329:22)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1218:10)
    at Module.load (internal/modules/cjs/loader.js:1047:32)
    at Function.Module._load (internal/modules/cjs/loader.js:935:14)
    at Module.require (internal/modules/cjs/loader.js:1087:19)
    at Module.require (pkg/prelude/bootstrap.js:1234:31)
    at require (internal/modules/cjs/helpers.js:73:18)
    at Object.<anonymous> (/snapshot/test/node_modules/sharp/lib/index.js:3:15)
    at Module._compile (pkg/prelude/bootstrap.js:1329:22)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1218:10)
    at Module.load (internal/modules/cjs/loader.js:1047:32)
    at Function.Module._load (internal/modules/cjs/loader.js:935:14)
    at Module.require (internal/modules/cjs/loader.js:1087:19)
    at Module.require (pkg/prelude/bootstrap.js:1234:31)
    at require (internal/modules/cjs/helpers.js:73:18)
```

I'm fairly convinced this has to do with #837 or #1066 or maybe a combination of both.

Looking at the below error message running in CI, it seems that prebuild-install is not correctly installed or referenced, this has been added in #1066:

```
> pkg@4.5.0
> Fetching base Node.js binaries to PKG_CACHE_PATH
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/sharp/build/Release
  %2: path-to-executable/sharp/build/Release
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/sharp/vendor/lib
  %2: path-to-executable/sharp/vendor/lib
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/sharp/build/Release
  %2: path-to-executable/sharp/build/Release
> Warning Cannot include directory %1 into executable.
  The directory must be distributed with executable as %2.
  %1: node_modules/sharp/vendor/lib
  %2: path-to-executable/sharp/vendor/lib
/bin/sh: 1: /builds/test/node_modules/pkg/node_modules/.bin/prebuild-install: not found
/bin/sh: 1: /builds/test/node_modules/pkg/node_modules/.bin/prebuild-install: not found
```

I'm mentioning both @david-mohr as well as @geekuillaume hoping they have a suggestion on how to fix this.

On a sidenote; this seems to me a breaking change so 5.0.0 would have been a more logical version for this release I suppose.

robertsLando:
When a `.node` file is required `pkg` copies it in a temporary folder like `/tmp/<hash>/addon.node` the problem is that for modules like `sharp` this isn't working as them are statically linked to other `.so` files using relative paths. This pr will copy the entire module folder inside the temporary directory, this allows also to remove the `tryImport` function as there is no reason to try again if that fails. 

I tested this on my end and it's working

I also grouped all hash folders inside `pkg` folder to be sure that the copy will run when users will update pkg, otherwise the copy may not be done when the folder already exists but was created with a previous pkg version

cc @jesec @leerob @erossignon 

Fixes #1075

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -2074,14 +2074,20 @@ function payloadFileSync(pointer) {
 
       // Node addon files and .so cannot be read with fs directly, they are loaded with process.dlopen which needs a filesystem path
       // we need to write the file somewhere on disk first and then load it
+      // the hash is needed to be sure we reload the module in case it changes
       const hash = createHash('sha256').update(moduleContent).digest('hex');
 
+      // Example: /tmp/<hash>
       const tmpFolder = path.join(tmpdir(), hash);
       if (!fs.existsSync(tmpFolder)) {
+        // here we copy all files from the snapshot module folder to temporary folder
+        // we keep the module folder structure to prevent issues with modules that are statically
+        // linked using relative paths (Fix #1075)
         fs.mkdirSync(tmpFolder, { recursive: true });
         copyFolderRecursiveSync(modulePkgFolder, tmpFolder);
       }
 
+      // replace the path with the new module path
       args[1] = path.join(tmpFolder, modulePackagePath, moduleBaseName);
     }
 
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -159,7 +159,7 @@ function createMountpoint(interior, exterior) {
 }
 
 function copyFileSync(source, target) {
-  var targetFile = target;
+  let targetFile = target;
 
   // If target is a directory, a new file with the same name will be created
   if (fs.existsSync(target)) {
@@ -172,10 +172,10 @@ function copyFileSync(source, target) {
 }
 
 function copyFolderRecursiveSync(source, target) {
-  var files = [];
+  let files = [];
 
   // Check if folder needs to be created or integrated
-  var targetFolder = path.join(target, path.basename(source));
+  const targetFolder = path.join(target, path.basename(source));
   if (!fs.existsSync(targetFolder)) {
     fs.mkdirSync(targetFolder);
   }
@@ -184,13 +184,20 @@ function copyFolderRecursiveSync(source, target) {
   if (fs.lstatSync(source).isDirectory()) {
     files = fs.readdirSync(source);
     files.forEach((file) => {
-      var curSource = path.join(source, file);
+      const curSource = path.join(source, file);
       if (fs.lstatSync(curSource).isDirectory()) {
         copyFolderRecursiveSync(curSource, targetFolder);
       } else {
         copyFileSync(curSource, targetFolder);
       }
     });
+  }
+}
+
+function createDirRecursively(dir) {
+  if (!fs.existsSync(dir)) {
+    createDirRecursively(path.join(dir, '..'));
+    fs.mkdirSync(dir);
   }
 }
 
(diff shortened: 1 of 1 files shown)


robertsLando:
@leerob Could you take a look into this please? If it looks ok to you could you also make a patch release? I need this in a project I'm working on, thannks :)

prelude/bootstrap.js
```diff
@@ -158,6 +158,42 @@ function createMountpoint(interior, exterior) {
   mountpoints.push({ interior, exterior });
 }
 
+function copyFileSync(source, target) {
+  var targetFile = target;
```
jesec, line 162: use `let`.

prelude/bootstrap.js
```diff
@@ -2020,88 +2056,41 @@ function payloadFileSync(pointer) {
...
       const moduleContent = fs.readFileSync(modulePath);
 
       // Node addon files and .so cannot be read with fs directly, they are loaded with process.dlopen which needs a filesystem path
       // we need to write the file somewhere on disk first and then load it
+      // the hash is needed to be sure we reload the module in case it changes
       const hash = createHash('sha256').update(moduleContent).digest('hex');
 
-      const tmpFolder = path.join(tmpdir(), hash);
+      // Example: /tmp/<hash>
+      const tmpFolder = path.join(tmpdir(), 'pkg', hash);
```
jesec, line 2081: tmpFolder would be `/tmp/pkg/<hash>` instead of `/tmp/<hash>`. Can you fix the comment?

prelude/bootstrap.js
```diff
@@ -2020,88 +2056,41 @@ function payloadFileSync(pointer) {
...
       if (!fs.existsSync(tmpFolder)) {
-        fs.mkdirSync(tmpFolder);
+        // here we copy all files from the snapshot module folder to temporary folder
+        // we keep the module folder structure to prevent issues with modules that are statically
+        // linked using relative paths (Fix #1075)
+        fs.mkdirSync(tmpFolder, { recursive: true });
```
jesec, line 2086: https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options

`recursive` is only supported on Node `v10.12.0` or later. We are still supporting Node v8 at the moment, and technically can't drop it without a major release.

robertsLando: Suggestions?

prelude/bootstrap.js
robertsLando: Sure

prelude/bootstrap.js
jesec: You would have to bake it yourself, like what you did with the `copyFolderRecursiveSync`. 

Side note: Node v16.7.0 or later has [fs.cpSync](https://nodejs.org/api/fs.html#fs_fs_cpsync_src_dest_options), which makes those custom functions unnecessary. However, v14 EOL in 2023.

leerob:
Thanks for the review, @jesec! 🙏

prelude/bootstrap.js
robertsLando: Check last commit

jesec (approved):

robertsLando:
@leerob new patch please 🙏🏼

ricreis394:
@Grewamor I'm having the same problem, and since I'm using yarn I don't know how can I install it to specific platform and arch, AFAIK yarn makes it impossible to choose

chuongmep:
Any update for now ? I'm facing with same issue
image

Issue #1291, opened 2021-08-17
ES modules not supported


LinusU:
I'm getting the following error as soon as the compiled app boots:

```text
node:internal/modules/cjs/loader:930
  throw err;
  ^

Error: Cannot find module '/snapshot/dhjaks/index.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:927:15)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1776:46)
    at Function.Module._load (node:internal/modules/cjs/loader:772:27)
    at Function.runMain (pkg/prelude/bootstrap.js:1804:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

Here is a minimal reproducible example:

**`package.json`**

```json
{ "type": "module" }
```

**`index.js`**

```js
import os from 'os'

console.log(os.arch())
```

Build command:

```sh
pkg index.js
```

sartaj-singh:
You didn't post your complete package.json file. So i think you didn't add your sub folders as scripts. For example if you want to add  dhjaks folder as script folder then you need to add this in package file like. 
{
  "name": "mdm5",
  "version": "1.0.1",
  "description": "MDM",
  "main": "start.js",
  "bin": "start.js",
  "scripts": {
    "start": "node ."
  },
  "pkg": {
    "scripts": [
      "dhjaks/*.js"
    ],
    "assets": [],
    "targets": [
      "node12",
      "linux-x64",
      "macos-x64",
      "win-x64"
    ]
  },
  "author": "demo",
  "license": "ISC",
  "dependencies": {
  }
}
assume dhjaks is subfolder under pacakge file parent folder. use build command pkg ./package.json

LinusU:
@sartaj-singh I actually did post my complete `package.json` file ☺️ 

The `dhjaks` folder is the folder of my entire package. My entire package only has two files: `package.json` & `index.js`. The goal is for it to just print out one line and then exit.

I did it like this to make a minimal test case that shows the problem. 

-----

I now tried to use `pkg package.json` instead:

```sh
$ mkdir foobar
$ cd foobar
$ echo '{ "name": "test", "bin": "index.js", "type": "module" }' > package.json
$ echo 'import os from "os"' > index.js
$ echo 'console.log(os.arch())' >> index.js
$ npx pkg package.json
$ ./test
node:internal/modules/cjs/loader:930
  throw err;
  ^

[...]

Issue #782, opened 2019-10-16
Support ES6 Modules


opensas:
I'm trying to package a node app as an exe using pkg, and I'd like to use ES6 imports.

I have something like this in my src/app.js:

```javascript
import express from 'express'
const app = express()

const eco = (req, res) => {
  const { method, url, headers, query } = req
  res.json({ method, url, headers, query })
}

app.all('/', eco)

app.listen(3000, () => console.log(`listening on http://localhost:3000`))
```

in my package.json I have:

```json
{
  "name": "pkg-test",
  "version": "1.0.0",
  "description": "",
  "main": "src/app.js",
  "type": "module",
  "type": "module",
  "scripts": {
"build": "pkg --targets=node12-win-x64 --output=iisnode-pkg.exe --options experimental-modules src/app.js",
    "start": "node --experimental-modules src/app.js",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "express": "^4.17.1"
  },
  "devDependencies": {
    "pkg": "^4.4.0"
  }

}
```

npm start works fine

```shell
$ npm start

> iisnode-pkg@1.0.0 start C:\data\devel\apps\tmp\iisnode-pkg
> node --experimental-modules src/app.js

(node:10668) ExperimentalWarning: The ESM module loader is experimental.
welcome to iisnode-pkg
iisnode-pkg listening on http://localhost:3000
```

but `npm run build` gives a waning and then running the exe throws an error:

```shell
>npm run build

> iisnode-pkg@1.0.0 build C:\data\devel\apps\tmp\iisnode-pkg
> pkg --targets=node12-win-x64 --output=iisnode-pkg.exe src/app.js --config package.json

> pkg@4.4.0
> Warning Failed to make bytecode node12-x64 for file C:\snapshot\iisnode-pkg\src\app.js
```

```shell
>iisnode-pkg.exe
C:\snapshot\iisnode-pkg\src\app.js:2
import express from 'express'
       ^^^^^^^

SyntaxError: Unexpected identifier
    at Module._compile (internal/modules/cjs/loader.js:701:23)
    at Module._compile (pkg/prelude/bootstrap.js:1268:32)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:768:10)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at Function.Module._load (internal/modules/cjs/loader.js:553:12)
[...]

PR #1323, opened 2021-09-30
feat: ESM modules support

robertsLando:
Ref #1291

To run test: `FLAVOR=test-00-esm npm test`

### DIFF
--- a/lib/follow.ts
+++ b/lib/follow.ts
@@ -4,6 +4,7 @@ import { core, sync, SyncOpts } from 'resolve';
 import assert from 'assert';
 import fs from 'fs';
 import path from 'path';
+import { CachedInputFileSystem, ResolverFactory } from 'enhanced-resolve';
 import { toNormalizedRealPath } from './common';
 
 Object.keys(core).forEach((key) => {
@@ -117,5 +118,42 @@ export function follow(x: string, opts: FollowOptions) {
         },
       })
     );
+  });
+}
+
+export function followESM(x: string, opts: FollowOptions) {
+  return new Promise<string>((resolve, reject) => {
+    const fileSystem = new CachedInputFileSystem(fs, 4000);
+
+    fileSystem.readFileSync = (file) => {
+      if (opts.ignoreFile && opts.ignoreFile === file) {
+        return Buffer.from(`{"main":"${PROOF}"}`);
+      }
+
+      if (opts.readFile) {
+        opts.readFile(file);
+      }
+
+      return fs.readFileSync(file);
+    }
+
+    const resolver = ResolverFactory.createResolver({
+      extensions: opts.extensions as string[],
+      fileSystem
+    });
+
+    const lookupStartPath :string = opts.basedir || __dirname;
+    const context = {};
+    const resolveContext = {};
+
+    resolver.resolve(context, lookupStartPath, x, resolveContext, (err, result) => {
+      if(!result || err) {
+        reject(err || Error(`Cannot resolve ${x}`))
+      } else {
+        resolve(result);
+      }
+    });
+
+    
   });
 }
--- a/lib/walker.ts
+++ b/lib/walker.ts
@@ -21,7 +21,7 @@ import {
   toNormalizedRealPath,
 } from './common';
 
-import { follow, natives } from './follow';
+import { follow, followESM, natives } from './follow';
 import { log, wasReported } from './log';
 import * as detector from './detector';
 import {
@@ -752,7 +752,7 @@ class Walker {
 
     const basedir = path.dirname(record.file);
     try {
-      newFile = await follow(derivative.alias, {
+      newFile = await followESM(derivative.alias, {
         basedir,
         // default is extensions: ['.js'], but
         // it is not enough because 'typos.json'
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@babel/parser": "7.13.13",
     "@babel/types": "7.13.12",
     "chalk": "^4.1.0",
+    "enhanced-resolve": "5.8.3",
     "escodegen": "^2.0.0",
     "fs-extra": "^9.1.0",
     "globby": "^11.0.3",
--- a/test/test-00-esm/main.cjs
+++ b/test/test-00-esm/main.cjs
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
+const utils = require('../utils.js');
+const os = require('os');
+
+// ignore this test if nodejs <= 10 , as recent version of PNPM do not support nodejs=10
+const MAJOR_VERSION = parseInt(process.version.match(/v([0-9]+)/)[1], 10);
+
+if (MAJOR_VERSION < 14) {
+  console.log(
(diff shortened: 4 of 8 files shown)


jesec (changes requested):
There is a (huge) binary in the commit.

robertsLando:
Sorry the test has failed and i forgot to delete it, when it works the vacoom automatically removes it

jesec:
> Sorry the test has fails and i forgot to delete it, when it works the vacoom automatically removes it.
> 
> Just to be clear tests will not pass as this is not a fix to #1291 but it just reproduce the problem, I tried to fix it without success. Do you have any clue what's not working?

#782 

We have no support for ES6 modules yet. The walker can't parse the dependencies and include them in the payload.

robertsLando:
The strange thing is that the detector actually identifies `import` statements: https://github.com/vercel/pkg/blob/59125d1820cdb380f3f592604bcfd7c017495e77/lib/detector.ts#L258

Also in latest ncc updates now it outputs esm modules so it's not a solution at all. I think we need to add a support to esm modules as with node 14+ them are supported. If you could point me on the changes needed I could try to fix this.

robertsLando:
@jesec news here?

robertsLando:
This could be a starting point: https://github.com/vitejs/vite/pull/5593

jesec:
Our use case is considerably more complicated. `enhanced-resolve` is nowhere near to a drop-in replacement for `resolve` in our case. 

We would have to study the APIs of the `enhanced-resolve` and rewrite our logics.

robertsLando:
> We would have to study the APIs of the enhanced-resolve and rewrite our logics.

I agree, in my tests I was just tring to see what happened using that as a separete tool then the plan it's to completely replace resolve. I think this issue should be the top in our todos, do you want to do some tries? Test is already setted up

hcldan:
I have been using es6 in my projects, but before I run pkg I transpile them to commonjs.  This seems to work well.  It would be great if this support landed natively or if you just built in the transpiling part so I didn't have to.

robertsLando:
> I transpile them to commonjs

Do you use webpack for this?

hcldan:
rollup worked really well, as I don't need the features of webpack, I just need my modules to be cjs in the end.

package.json
```json
{
  "scripts": {
     "build": "npm-run-all build:*",
     "build:1": "rimraf dist",
     "build:2": "mkdirp dist",
     "build:4": "rollup index.js -o dist/built.js --format cjs",
     "build:5": "pkg dist/built.js --targets linux,win --out-path dist"
  }
}
````

**Comment by hcldan** [comment]
sourcemaps are a pain... so native support would be fantastic

**Comment by rightaway** [comment]
I can help to test when it is ready.

**Comment by RUGMJ7443** [comment]
are there any updates or workarounds for this? I really need to distrabute my package ASAP

**Comment by kamack38** [comment]
> are there any updates or workarounds for this? I really need to distrabute my package ASAP

You can use `caxa`, because they are supporting ESM.

**Comment by hcldan** [comment]
I've commented above for a decent workaround that I'm using

rollup can be configured to transpile and bundle all es modules in a file and defer all commonjs modules to the default require.
```

  "scripts": {
    "test": "npm-run-all test:*",
    "test:1": "eslint",
    "test:2": "NODE_OPTIONS=--experimental-vm-modules npx jest",
    "start": "cross-env node .",
    "build": "npm-run-all build:*",
    "build:1": "rimraf dist",
    "build:2": "mkdirp dist",
    "build:4": "rollup --config rollup.config.js",
    "build:5": "pkg dist/nwsp.js --targets linux,win --out-path dist"
  },
```

rollup.config.js
```
import nodeResolve from 'rollup-plugin-node-resolve';

export default {
  input: 'index.js',
  output: {
    file: 'dist/nwsp.js',
    format: 'cjs',
  },
  plugins: [
    nodeResolve({
      modulesOnly: true,
    }),
  ],
};
```

There are tons of options, look them over.  This gives you one big file with all es5 turned into commonjs.
Only big drawbacks here are complexity (it would be nice if this all just worked) and sourcemap support.
Rollup can output sourcemaps (even inline) but I cant get stacks reported by node to use that info.

**Comment by gaby** [comment]
Any updates on this feature?

**Comment by rightaway** [comment]
Would also like to know what the plans are with ESM modules support. Some dependencies like https://github.com/sindresorhus/got only work with ESM modules now and encourage switching (https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

The options are to stay on older versions of dependencies or to stop using pkg but it's better if we don't have to do either.

**Comment by robertsLando** [comment]
This PR doesn't fixes this, @jesec has a working implementation but not finished yet and I dunno if there is any ETA on it

**Comment by gaby** [comment]
Any reason why stale bot is trying to close PR's?

**Comment by CleyFaye** [comment]
> Any reason why stale bot is trying to close PR's?

I'd say it's

> because it has been open 90 days with no activity

**Comment by angrymouse** [comment]
Any updates on it?

**Comment by edosrecki** [comment]
Any news at all?

**Comment by cspotcode** [comment]
I think there is another issue here: even if the files are bundled into the virtual filesystem, it cannot be executed.  I'm guessing node's ESM resolver does not use the patched filesystem.

For example, you can bundle an ESM file alongside an CJS file, using the CJS file as entrypoint.  Then run stuff in the CJS file that looks vaguely like this:

```
// entrypoint.cjs:
// Prove that the file exists in VFS
console.log(fs.readdirSync('/snapshot'));
console.log(fs.readFileSync('/snapshot/bin.mjs'));
// But import fails
import('/snapshot/bin.mjs');
```

Gives a cryptic error:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/snapshot/bin.mjs' imported from /snapshot/entrypoint.cjs
Did you mean to import /snapshot/bin.mjs?
```

Looking closely at that error, it says i cannot find the module, then it suggests exactly the same path.  I bet the suggestion comes from different code which is using the VFS.

robertsLando:
Unfortunately I have no clue about how to actually implement such feature, I would need some support on this

Haringat:
@robertsLando Perhaps you could create a custom loader that uses the virtual filesystem.
Here are some example implementations from the node docs: https://nodejs.org/dist/latest-v19.x/docs/api/esm.html#loaders

robertsLando:
@Haringat Thanks for your suggestion, BTW I don't think that would solve the issue, right now the problem is we don't have a way to resolve the modules used with esm, enhanced resolve should solve this issue but I have no clue how to make it work

Haringat:
@robertsLando But with the loader you can specify a custom resolve hook to do just that. It gets the URL it is supposed to resolve and the URL of the importing module.

robertsLando:
It would not work for dynamically loaded modules I think? I mean an async import for example will not be catched by it on startup

Haringat:
@robertsLando I just tested it with an HTTP/HTTPS loader and it was also invoked for a dynamic import.

robertsLando:
The problem is different here, we actually don't run a file with node to detect it's dependencies, we start from the entrypoint file and from that we recursively resolve all modules using "resolve" npm module, to detect them we simply use regex to find require/import statements in the files. Your suggestion so would not work in this case at is requires to run the entrypoint file, also there is no way to know how to trigger dynamic import in that way.

PR #1324, opened 2021-09-30
fix: catch babel parse errors

### LINKED ISSUE #746: [bug] Ignore .vue files inside node_modules
I'm getting this error

```
> Error! Unexpected token (3:0)
  /myApp/node_modules/vue2-google-maps/dist/components/infoWindow.vue
```

Seems that pkg is trying to compile a vue files and this throws the error shown. How can  this be avoided?

robertsLando:
Fixes #746

### DIFF
--- a/lib/detector.ts
+++ b/lib/detector.ts
@@ -7,7 +7,6 @@ import * as babelTypes from '@babel/types';
 import * as babel from '@babel/parser';
 import { log } from './log';
 
-
 import { ALIAS_AS_RELATIVE, ALIAS_AS_RESOLVABLE } from './common';
 
 function isLiteral(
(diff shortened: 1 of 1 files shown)


robertsLando:
> We should find out why pkg tries to parse a .vue file.

I haven't been able to detect where this happens, maybe because It try to parse assets somewhere, BTW this could be a temporary fix until we find a better solution?

jesec:
I am worried about the risk that the real errors got ignored.

robertsLando:
Maybe I could add a log so at least users know something has fail?

jesec:
Yeah. We can downgrade it (from error) to warning for the time being.

robertsLando:
@jesec Could we also merge this?

robertsLando:
Strange the eslint didn't catch them on CI (and also on my editor) 😕 BTW it's fixed now

PR #1327, opened 2021-10-04
bootstrap: skip regex when dlopen path isn't inside snapshot

### LINKED ISSUE #1319: cross platform native modules didn't work for me
re: https://github.com/vercel/pkg/pull/837 - is cross-platform native module packing supposed to work, or not? Prebuild-install is supposed to do this right: https://github.com/vercel/pkg/blob/main/lib/producer.ts#L200

pkg 5.3.2
souce os: macos 10.15.7
target: win
how am I configuring pkg? Just with cmd parameters in my package.json: `pkg src/my.js -t win --out-path /Volumes/[C]/bin/myname` (I tried specifying the target with node version and architecture, made no difference)

# What happened?
pkg packaged up the macos version of iconv's module: iconv.node into the windows package. When I ran the exe, Windows errored this wasn't a win32 application

# What did I expect to happen?
From the readme or `--help` I wasn't sure: Either to bundle the correct file, or If the target os's native module isn't available, some automatic way to fall back to the old behaviour of expecting the right .node file to be available after compile in the location of the binary, failing that, a config setting that will give me the old behaviour or allow me to bundle the correct .node file manually (does specifying an asset override native module compilation, or not?). Currently I don't get how to achieve any of those (since i've been best with the post-compile same-location method until now)

# How did I temporarily/manually fix my problem?
I copied the windows native module iconv.node into the location in mac's node_modules where the mac inconv.node should be, then ran the packaging again, then once the binary was made, I copied the mac iconv.node back where it should be.

How can I automate that better, or just get the old behaviour back (easiest is just how do I stop pkg trying to bundle the .node file here?)

Also how can we better-document what the correct approach is to cross-platform native module compilation?

robertsLando:
When tring to import a file with `dlopen` and it's not inside snapshot regex could fail and throw error

Fixes #1319

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -2064,19 +2064,19 @@ function payloadFileSync(pointer) {
     const moduleBaseName = path.basename(modulePath);
     const moduleFolder = path.dirname(modulePath);
 
-    // Example: moduleFolder = /snapshot/appname/node_modules/sharp/build/Release
-    const modulePkgPathRegex = /.*?node_modules\/((.+?)\/.*)/;
-    // Example: modulePackagePath = sharp/build/Release
-    const modulePackagePath = moduleFolder.match(modulePkgPathRegex)[1];
-    // Example: modulePackageName =  sharp
-    const modulePackageName = moduleFolder.match(modulePkgPathRegex)[2];
-    // Example: modulePkgFolder = /snapshot/appname/node_modules/sharp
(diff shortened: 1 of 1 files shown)


jesec (approved):
lg2m

robertsLando:
@leerob I don't need it as my project is working even without this BTW a patch release could be useful to users facing #1319  🙏🏼

PR #1344, opened 2021-10-22
feat: add an option for turning off Node.js hijacking

### LINKED ISSUE #897: Error: Command failed: Executing pkg cli from the child process of pkg cli getting failed.
We have a requirement to call the shell script from pkg CLI and that script is again calling the same pkg CLI.

Our CLI is calling the shell script using the child_process package of the node. In a shell script, we are again calling the Pkg CLI if some condition is met. 

Following is the testing Code: test.js

```
const doSomething1 = () => {
  const childProcess = require('child_process');
  console.log('Do Something 1 called');
  // calling some shell script. and in script we have some condition if it is true then call doSomething2 from executables.
  childProcess.execSync('./pkgBinary 2');
}

const doSomething2 = () => {
  console.log('Do Somthing 2 called');
}

if (process.argv[2] === '1') {
  doSomething1();
} else {
  doSomething2();
}
```

Created binary using  

`pkg test.js -t 'node12-macos-x64' -o 'pkgBinary'`

after the execution of command './pkgBinary 1'. I am getting the following error.

Error: Command failed: ./pkgBinary 2
internal/modules/cjs/loader.js:983
  throw err;
  ^

Error: Cannot find module '/Users/appirio-13532/practice/2'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:980:15)
    at Function.Module._resolveFilename (pkg/prelude/bootstrap.js:1346:46)
    at Function.Module._load (internal/modules/cjs/loader.js:862:27)
    at Function.Module.runMain (pkg/prelude/bootstrap.js:1375:12)
    at internal/main/run_main_module.js:18:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

    at checkExecSyncError (child_process.js:630:11)
    at Object.execSync (child_process.js:666:15)
    at Object.childProcess.execSync (pkg/prelude/bootstrap.js:1507:30)
    at doSomething1 (/snapshot/practice/test.js)
    at Object.<anonymous> (/snapshot/practice/test.js)
    at Module._compile (pkg/prelude/bootstrap.js:1320:22)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.Module.runMain (pkg/prelude/bootstrap.js:1375:12) {
  status: 1,
  signal: null,
  output: [
    null,
    <Buffer >,
    <Buffer 69 6e 74 65 72 6e 61 6c 2f 6d 6f 64 75 6c 65 73 2f 63 6a 73 2f 6c 6f 61 64 65 72 2e 6a 73 3a 39 38 33 0a 20 20 74 68 72 6f 77 20 65 72 72 3b 0a 20 20 ... 450 more bytes>
  ],
  pid: 14642,
  stdout: <Buffer >,
  stderr: <Buffer 69 6e 74 65 72 6e 61 6c 2f 6d 6f 64 75 6c 65 73 2f 63 6a 73 2f 6c 6f 61 64 65 72 2e 6a 73 3a 39 38 33 0a 20 20 74 68 72 6f 77 20 65 72 72 3b 0a 20 20 ... 450 more bytes>
}

zkochan:
close #897

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -680,7 +680,8 @@ export async function exec(argv2: string[]) {
       target: target as Target,
       symLinks,
       doCompress,
-      nativeBuild
+      nativeBuild,
+      mockNode: process.env.MOCK_NODE !== 'false'
     });
 
     if (target.platform !== 'win' && target.output) {
--- a/lib/packer.ts
+++ b/lib/packer.ts
@@ -159,7 +159,7 @@ export default function packer({
     }
   }
   const prelude =
-    `return (function (REQUIRE_COMMON, VIRTUAL_FILESYSTEM, DEFAULT_ENTRYPOINT, SYMLINKS, DICT, DOCOMPRESS) { 
+    `return (function (REQUIRE_COMMON, VIRTUAL_FILESYSTEM, DEFAULT_ENTRYPOINT, SYMLINKS, DICT, DOCOMPRESS, MOCK_NODE) { 
         ${bootstrapText}${
       log.debugMode ? diagnosticText : ''
     }\n})(function (exports) {\n${commonText}\n},\n` +
@@ -172,6 +172,8 @@ export default function packer({
     '%DICT%' +
     '\n,\n' +
     '%DOCOMPRESS%' +
+    '\n,\n' +
+    '%MOCK_NODE%' +
     `\n);`;
 
   return { prelude, entrypoint, stripes };
--- a/lib/producer.ts
+++ b/lib/producer.ts
@@ -252,6 +252,7 @@ interface ProducerOptions {
   symLinks: SymLinks;
   doCompress: CompressType;
   nativeBuild: boolean;
+  mockNode: boolean;
 }
 
 /**
@@ -316,6 +317,7 @@ export default function producer({
   symLinks,
   doCompress,
   nativeBuild,
+  mockNode,
 }: ProducerOptions) {
   return new Promise<void>((resolve, reject) => {
     if (!Buffer.alloc) {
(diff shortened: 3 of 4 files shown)


zkochan:
Does it mean that the issue #897 can be solved by simply running `PKG_EXECPATH=PKG_INVOKE_NODEJS pkg app.js` ?

jesec:
> Does it mean that the issue #897 can be solved by simply running `PKG_EXECPATH=PKG_INVOKE_NODEJS pkg app.js` ?

If you run a pkg executable with `PKG_EXECPATH=PKG_INVOKE_NODEJS`, it would behave the same as a plain `node`. That may or may not be the behavior you want.

zkochan:
Why was it closed? Is #897 fixed?

robertsLando:
Like @jesec said you should get the same result using `PKG_EXECPATH` env var set to `PKG_INVOKE_NODEJS`

zkochan:
For sure not the same result because see my change at line 1919

robertsLando:
You should use an env var so to make it configurable

zkochan:
ok

zkochan:
This is not doing what I need.

This solution works if I set the env variable, when running the compiled binary. What I want is to build the binary with this setting on.

zkochan:
Probably I need to add a new option here

https://github.com/vercel/pkg/blob/f081c634c73aaf00cff69750c561771442092098/lib/packer.ts#L161-L175

zkochan:
Could someone look into this please?

robertsLando (approved):
Looks good from a code side. Dunno if @jesec agree

edvald:
Just FYI, here's how we solved this for ourselves with a patch: https://github.com/vercel/pkg/issues/897#issuecomment-1049370335

For clarity, the exact issue for us is the patching of `child_process`. We needed a way to explicitly disable part of it (at least in lieu of a more general solution which I couldn't come up with myself), so that e.g. scripts run in child processes created by a pkg'd binary could successfully call the parent pkg'd binary. The goal is not to run node directly, which I believe is already sorted with the `PKG_EXECPATH=PKG_INVOKE_NODEJS` option.

PR #1393, opened 2021-11-11
ldid: explicitly mark signature as adhoc

asdfugil:


jesec:
Can you explain why this is neccessary?

asdfugil:
it is needed for Monterey

jesec (approved):
I am able to run the binary (produced with the current flags) on macOS Monterey. That is, this PR is not really neccessary.

```
Executable=/Users/jc/Sources/pkg/1
Identifier=pkg-macos
Format=Mach-O thin (arm64)
CodeDirectory v=20400 size=389954 flags=0x0(none) hashes=12181+2 location=embedded
./1: no signature
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements count=1 size=128
```

Nonetheless, I am happy to make the result closer to the codesign utility.

asdfugil:
sorry i mixed it up with iOS 15 (which we don't target), and that doesn't allow JIT anyways so this really isn't necessary

nathan818fr:
What version of ldid are you using?
There is no `-C` option: https://github.com/xerub/ldid/blob/master/ldid.cpp
Since ldid terminates silently in case of unknown options, this change only skips the code-signing.

jesec:
https://github.com/jesec/ldid-static

PR #1435, opened 2021-12-10
fix: usage of .node files not in node_modules

paul-marechal:
Make sure to copy the .node file in the temporary folder before trying
to dlopen it.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -2116,7 +2116,10 @@ function payloadFileSync(pointer) {
         // Example: /tmp/pkg/<hash>/sharp/build/Release/sharp.node
         newPath = path.join(tmpFolder, modulePackagePath, moduleBaseName);
       } else {
-        // simple load the file in the temporary folder
+        createDirRecursively(tmpFolder);
+        copyFileSync(modulePath, tmpFolder);
+
+        // load the copied file in the temporary folder
         newPath = path.join(tmpFolder, moduleBaseName);
       }
 


robertsLando (approved):

robertsLando:
cc @leerob @jesec

jesec (changes requested):
I am confused. The actual change seems to be unrelated to the description. 

It is better described as "don't create folder and copy files, if the folder already exists". Additionally, changes to L2092-2099 are no-op, and the comment change needs extra justification.

paul-marechal:
> I am confused.

I was too... At some point the logic used to unconditionally "create some temp folder then move `.node` files then `dlopen` it".

A patch added the branching to check if the `.node` file is in some `node_modules` folder and properly create the file/folder hierarchy. In the `else` branch (`.node` file is not under some `node_modules` folder) the logic assumed the `.node` files could be `dlopen`'d but it is just not the case because the copy never happened in that branch.

I encountered this bug because I used Webpack to bundle an app nicely to make it easier when using pkg. Except part of the bundling moved the `.node` files under some `<project>/native` folder, hence why I was having issues with the `else` logic I proposed a fix for here.

About the comment: As far as I understand static linking means the shared library already contains its own dependencies, only dynamic linking cares about files relative to the binary. So the comment seemed wrong to me as I was trying to understand the logic.

jesec (approved):

jesec:
I dropped the no-op bits, and the rest looks good to me.

prelude/bootstrap.js
```diff
@@ -2116,7 +2116,10 @@ function payloadFileSync(pointer) {
         // Example: /tmp/pkg/<hash>/sharp/build/Release/sharp.node
         newPath = path.join(tmpFolder, modulePackagePath, moduleBaseName);
       } else {
-        // simple load the file in the temporary folder
+        createDirRecursively(tmpFolder);
+        copyFileSync(modulePath, tmpFolder);
```
paul-marechal, line 2120: Just out of curiosity: there used to be an `if` check to see if the file already existed, is this not necessary anymore?

jesec: See #1492.

Issue #1469, opened 2022-01-07
There will be an infinite loop, with only different dirname symLinks, when findCommonJunctionPoint, in windows OS


ranforce:
### What version of pkg are you using?

pkg@5.5.1

### Which version(s) of pkg work for you?

pkg@4

### What version of Node.js are you using?

v12.22.8

### What operating system are you using?

Windows 7 and 10 both

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

default,as 12.22.7-win-x64 in console

### Describe the Bug

Hello, in code https://github.com/vercel/pkg/blob/f081c634c73aaf00cff69750c561771442092098/lib/walker.ts#L306 , when the scope is as follows, there will be an infinite loop.
{
"file": "D:\\project-name\\node_modules\\element-ui\\package.json",
"realFile": "D:\\project-name\\node_modules\\_element- ui@2.15.5 @element-ui\\package. json"
}
The final scope is:
{
"file": "D:\\",
"realFile": "D:\\"
}

### Expected Behavior

Build sussece

### To Reproduce

The scope only the dirname is diffrent.
{
"file": "D:\\project-name\\node_modules\\element-ui\\package.json",
"realFile": "D:\\project-name\\node_modules\\_element- ui@2.15.5 @element-ui\\package. json"
}

robertsLando:
PR?

awolokita:
I have also encountered this issue and would like to add a few more observations.

- This issue is encountered when running `pkg` as a part of our Jenkins CI job.
- The package target is a lerna monorepo sub-package with dependencies on other packages in the mono-repo
- Only encountered when running with the `sh` Jenkins command (uses Git Bash).
- When running with the `powershell` command the issue is not seen. NB: lerna bootstrap and tsc must both be run with `powershell` in order to work around the issue.
- Moreover, using git bash (even with CI=true) manually on the Jenkins node doesn't cause this issue. Issue is also not seen when using 
- When this issue is seen, `findCommonJunctionPoint` gets stuck in the while-loop, continuously moving up directories until it reaches `C:\` and gets stuck there.
- I added additional `consol.log` in the `findCommonJunctionPoint` loop to output the current `file` and `realFile`: it looks like `file === `"c:\\"` while `realFile === "C:\\"`.
- So possibly the issue is that for some reason, the difference in case of the root is the cause of the issue?

[...]

PR #1470, opened 2022-01-07
fix infinite loop issue #1469

ranforce:
issues:
https://github.com/vercel/pkg/issues/1469
when the scope is as follows, there will be an infinite loop.
{
"file": "D:\project-name\node_modules\element-ui\package.json",
"realFile": "D:\project-name\node_modules\_element- ui@2.15.5 @element-ui\package. json"
}
The final scope is:
{
"file": "D:\",
"realFile": "D:\"
}


jesec:
Thanks for the PR. However, it seems to cause the tests to fail.

PR #1484, opened 2022-01-17
fix: copyFile and copyFileSync patch

### LINKED ISSUE #420: fs.copyFile and snapshot filesystem
Trying to copy a file from the snapshot filesystem to the real filesystem results in the following error:
`ENOENT: no such file or directory`. Seems to happen in all versions after 4.2.4.

Example code that should work:
```javascript
fs.copyFile(
  path.resolve(__dirname, 'some_file.json'),
  path.resolve(process.cwd(), 'some_new_file.json')
)
```

robertsLando:
Fixes #420

Based on @medikoo solution: https://github.com/vercel/pkg/issues/420#issuecomment-652925540


robertsLando:
@jesec Ping please

robertsLando:
Anyone that could look at this? 😢  @jesec

jesec:
looks OK. but I don't think it is a clever idea to `readFile` and then `writeFile`. Is it possible to do it with stream or chunks? 

Also is the function for extracting native addons still required? Can't we just use this?

jesec (approved):

jesec:
There appear to be some weird issues with ESLint. 

Also, we may want to do the same thing for the new `fs.cp` and `fs.cpSync`. 

https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fscpsyncsrc-dest-options

PR #1492, opened 2022-01-27
Copy module folders always to temporary folder

trippier1:
We have an issue reported by several users showing error message “Error: The specified module could not be found.” where some .node files cannot be found in Temp folder under Windows. Because of this error, application cannot be started.
After analysis we have seen that the directories inside temp remain but files are cleaned up (we don’t know which clean up mechanism, Windows OS or antivirus software but we anyway have no means to intervene here). pkg currently only checks for existence of the folder pkg/<hash> and if exists, no copying is done. Since one can never be sure about what happens with files and directories in temporary folder when the application is not active, safer way would be to always do the copying.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -2091,13 +2091,11 @@ function payloadFileSync(pointer) {
         // Example: modulePkgFolder = /snapshot/appname/node_modules/sharp
         const modulePkgFolder = parts.slice(0, mIndex + 1).join(path.sep);
 
-        if (!fs.existsSync(tmpFolder)) {
-          // here we copy all files from the snapshot module folder to temporary folder
-          // we keep the module folder structure to prevent issues with modules that are statically
-          // linked using relative paths (Fix #1075)
-          createDirRecursively(tmpFolder);
-          copyFolderRecursiveSync(modulePkgFolder, tmpFolder);
-        }
+        // here we copy all files from the snapshot module folder to temporary folder
+        // we keep the module folder structure to prevent issues with modules that are statically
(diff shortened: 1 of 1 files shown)


btsimonh:
Note: I don't know where the files go in pkg, but in my independent implementation of something similar, I now extract the files to folders other than temp.  I never had an automatic cleanup issue on Windows, but did on both linux and OSX.  I now use the 'programdata' folder equivalents on all platforms to avoid this.

jesec (approved):

trippier1:
@jesec thanks for the review. Can you please also approve running workflow?

robertsLando (approved):

robertsLando:
@jesec I think this is good to merge

robertsLando:
@trippier1 Another possible fix would be to allow providing tmp folder using an env var like `PKG_TMP_FOLDER`. This should allow users to place node addons files where they prefer

jesec:
> @trippier1 Another possible fix would be to allow providing tmp folder using an env var like `PKG_TMP_FOLDER`. This should allow users to place node addons files where they prefer

Ugh. We don't need more options, especially not the runtime env variables.

jesec:
A CI run failed. I requeued it. Let's see how things go.

robertsLando:
I don't mean a runtime env var, it could be just an env var user can set from executable if he wants

jesec:
> I don't mean a runtime env var that we need to bake into the executable, it could be just an env var user can set from executable if he wants

That's even worse. A baked in option is for the users, while a runtime env variable is for the end-users. We have to be much much more careful on options to provide to the end-users. So absolutely no.

PR #1532, opened 2022-02-20
fix: copy file from snapshot if spawned

### LINKED ISSUE #1516: ENOENT when executing bundled binary asset
### What version of pkg are you using?

5.5.1

### What version of Node.js are you using?

16.13.1

### What operating system are you using?

Windows

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

latest-win

### Describe the Bug

I'm trying to spawn a bundled .exe, but I get ENOENT.
I'm including the bin folder:
```
"pkg": {
    "assets": ["bin"]
  }
```
And in my code I use the following (as instructed by the pkg readme):
```
let bin = path.join(__dirname, '../bin/pg_dump.exe')
spawn(bin, ['--version'])
```
Also there are dll's in the bin folder required to run pg_dump.exe

### Expected Behavior

To spawn the bundled .exe

### To Reproduce

Try to bundle an .exe and execute it

kldzj:
This change wraps the spawn internals to copy files from snapshot if needed.

The approach has a few downsides, if in fact the executable needs to read other snapshot files, it will obviously fail. For such advanced use cases copying manually still is a viable option. Perhaps this is undesired behaviour (e.g. creating tmpdirs in pkg's name).

Fixes #1516


robertsLando (approved):

prelude/bootstrap.js
```diff
@@ -2029,6 +2029,66 @@ function payloadFileSync(pointer) {
...
+        if (shouldRunUsingCmd(file)) {
+          options.file = 'cmd.exe';
+          options.args = ['/c', newFile, ...options.args];
+        } else {
+          options.file = newFile;
+        }
+
+        fs.copyFileSync(file, newFile);
+        if (process.platform !== 'win32') {
+          fs.chmodSync(newFile, fs.constants.S_IRWXU);
```
jesec, line 2071: The snapshot recorded the original permissions of the file. We should restore that instead.

prelude/bootstrap.js
```diff
@@ -2029,6 +2029,66 @@ function payloadFileSync(pointer) {
...
+          }
+        }
+      }
+
+      function shouldRunUsingCmd(file) {
+        return (
+          process.platform === 'win32' &&
+          (file.endsWith('.cmd') || file.endsWith('.bat'))
+        );
+      }
```
jesec, line 2054: Is this really necessary? 

Shouldn't the users do this themselves?

jesec (changes requested):
See comments. 

Also, in general, I don't like the idea. The use case is very limited and weird. I am not sure that if we should support it. 

Another thing to note is that, more often than not, the executables (e.g. `.exe`) could depend on libraries (e.g. `.dll`) residing in the same directory. This PR does not handle such case, and IMO it would be too complicated (and certainly require config fields) if that has to be implemented. 

I think the better course of action would be to ask the users to add routines themselves to extract executables to temporary folders, now we have implemented `snapshot -> actual` copyFile.

prelude/bootstrap.js
kldzj: Without it it won't be running on windows, I think it's an even worse idea to search for a path in the args.

kldzj:
@jesec Please feel free to close this PR if you feel this should not be supported by pkg ootb. Otherwise I'll implement the changes you asked for. 👍🏻

kldzj:
Regarding your concerns, while the use case is limited I don't think it's weird. 

As you and my PR mentioned, if the executable requires reading other snapshot files it will fail to do so and the developer should copy whatever is necessary to the real fs and spawn from there. But for simply running scripts or other binaries that don't need to read from the snapshot it saves the hassle of writing the copy and cleanup logic.

kldzj:
Hey @robertsLando or @jesec, can you please re-run `14-windows`? It ran fine on [my branch](https://github.com/kldzj/pkg/runs/5342117897?check_suite_focus=true).

robertsLando:
@kldzj done

kldzj:
I'm unsure why running [this pkg command](https://github.com/vercel/pkg/blob/main/test/test-775/main.js#L46-L49) would (sometimes) fail since my last commit. Why would pkg not have permission to open the output file a second time?

robertsLando:
@kldzj  no clue, I'm sorry. @jesec ?

jesec:
> @kldzj no clue, I'm sorry. @jesec ?

I think there is something wrong with the virtual environment of GitHub Actions.

Meanwhile, I still don't like the idea. However, I am giving some thought to a general mechanism to allow users to extract some files from the snapshot. I am tied up with some other matters at the moment, and I will make the decision when I have more time.

Gu7z:
I have a case like @janpio. A cli that should run some tasks in Windows and Linux. There is a pre-builded application for each SO (.exe | .AppImage) but the idea now is merge both into a single CLI based on the old ones. It'll be nice to have the option to build executable with pkg. As @kldzj mentioned, yes is a limited case but not necessary weird.

kldzj:
Not stale.

robertsLando:
cc @jesec

PR #1547, opened 2022-03-09
fix: do not overwrite temp .node files

paul-marechal:
This can happen when two processes try to load the same .node file.


paul-marechal:
The `CI / test` failures seem unrelated?

jesec:
See #1492 .

paul-marechal:
@jesec I know about https://github.com/vercel/pkg/pull/1492, and there the issue happened because it used to check for the _folder_ to exist, and I can anecdotally say that I also encountered the same issue in the past where the folder would exist but the `.node` file would be missing.

This PR is different because it checks for the file's existence.

robertsLando (approved):
@jesec I agree, the other PR is for the first if case where the entire folder is copied

jesec (approved):

robertsLando:
Can this be merged?

PR #1594, opened 2022-04-19
CI: Add Node 18 to version matrix

stscoundrel:
Node 18 became "current" Node.js version today. Add it to version matrix in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false # prevent test to stop if one fails
       matrix:
-        node-version: [10.x, 12.x, 14.x, 16.x]
+        node-version: [10.x, 12.x, 14.x, 16.x, 18.x]
         os: [ubuntu-latest, windows-latest] # Skip macos-latest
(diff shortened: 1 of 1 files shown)


robertsLando (approved):

stscoundrel:
The workflow seems to have failed, because Github Actions did not yet have version of Node 18 available in "18" syntax. It works fine with "18.x" syntax, which defaults to build from nodejs.org if one is not found.

They seem to have added 18.0.0 yesterday -> It should work now if you rerun the workflows. I don't seem to have privileges to rerun them.

jesec:
This can't be merged before we have (patched) Node 18 binaries in pkg-fetch.

jesec:
> > This can't be merged before we have (patched) Node 18 binaries in pkg-fetch.
> 
> nodejs18 seems to have a new user snapshot API, consider using it?

That would not be useful for this project at this stage. No one can be happy with compiling Node binaries of their own (takes half an hour on a decent machine). 

I am closing this PR since it can't be merged. I will add Node 18 to the matrix when pkg-fetch is ready.

Issue #1589, opened 2022-04-14
Error: EBUSY: resource busy or locked on native modules


mertushka:
### What version of pkg are you using?

5.6.0

### Which version(s) of pkg work for you?

< 5.6.0

### What version of Node.js are you using?

16.14.0

### What operating system are you using?

Windows

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

node14-win-x64

### Describe the Bug

![unknown](https://user-images.githubusercontent.com/34413473/163461496-2cb91743-65c3-4671-b5ea-9e8ca622b1e6.png)

### Expected Behavior

Native modules must import succesfully

### To Reproduce

Idk.

marcinadrem:
I experienced the same behavior when executing multiple instances of packed executable. It seems like native module is locked in "FILE LOCKED WITH ONLY READERS" mode and second instance tries to open it in write mode. Please see attached log.

[Logfile.CSV](https://github.com/vercel/pkg/files/8541015/Logfile.CSV)

Please unpack attached file and run run_test.bat to reproduce

[load_native_module_test.zip](https://github.com/vercel/pkg/files/8541431/load_native_module_test.zip)

renkei:
I'm not sure but this could be related to [PR1492](https://github.com/vercel/pkg/pull/1492) in 5.6.0. The .node files are always copied now to make sure that the temp folder contains all related files, see PR for details. But if there is an already running instance the native module files are already opened and therefore cannot be overwritten. Solution could be to perform always the copy action, just to make sure that the tmp folder is always properly populated, but to allow to fail. Then the second instance of the pkg app should just proceed with the already existing file.

For instance, copyFolderRecursiveSync in line 183 of `prelude/bootstrap.js` could be modified like this:
```js
function copyFolderRecursiveSync(source, target) {
  let files = [];

  // Check if folder needs to be created or integrated
  const targetFolder = path.join(target, path.basename(source));
  if (!fs.existsSync(targetFolder)) {
    fs.mkdirSync(targetFolder);
  }

  // Copy
  if (fs.lstatSync(source).isDirectory()) {
    files = fs.readdirSync(source);
    files.forEach((file) => {
      const curSource = path.join(source, file);
      if (fs.lstatSync(curSource).isDirectory()) {
        copyFolderRecursiveSync(curSource, targetFolder);
      } else {
        try {
          fs.copyFileSync(
            curSource,
            path.join(targetFolder, path.basename(curSource))
[...]

PR #1611, opened 2022-05-01
fix(boootstrap): prevent to override existing node addon file

renkei:
This PR fixes [#1589](https://github.com/vercel/pkg/issues/1589). The fix has two effects:

1. `.node` files that must exist on the real filesystem instead of the virtual snapshot filesystem (because of process.dlopen) are only copied if they do not exist. If they already exist then it doesn't matter anymore whether they are currently opened by an running instance of the pkg-packaged app or not, because existing files are untouched now.
2. The start of the pkg app a second, third, etc. time is as fast as with pkg 5.5.2 again since (in best case) no copy action takes place

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -230,7 +230,7 @@ function copyFolderRecursiveSync(source, target) {
           // If checksums are equal then there is nothing to do here
           // ==> target already exists and is up-to-date
           if (curSourceHash === curTargetHash) {
-            return;
+            continue;
           }
         }
 
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -198,6 +198,12 @@ function copyFolderRecursiveSync(source, target) {
         copyFolderRecursiveSync(curSource, targetFolder);
       } else {
         const curTarget = path.join(targetFolder, path.basename(curSource));
+        // Check if the target file already exists and skip the copy in such a case to
+        // 1. avoid an exception that the file cannot be written if it is already opened
+        //    by a running instance of the pkg-packaged app
+        // 2. speed up the second, third, etc. app start if (in best case) all files
+        //    are already existing
+        // See https://github.com/vercel/pkg/issues/1589 for more details.
         if (!fs.existsSync(curTarget)) {
           fs.copyFileSync(curSource, curTarget);
         }
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -181,34 +181,71 @@ function copyInChunks(
 
 // TODO: replace this with fs.cpSync when we drop Node < 16
 function copyFolderRecursiveSync(source, target) {
-  let files = [];
-
-  // Check if folder needs to be created or integrated
+  // Build target folder
   const targetFolder = path.join(target, path.basename(source));
+
+  // Check if target folder needs to be created or integrated
   if (!fs.existsSync(targetFolder)) {
     fs.mkdirSync(targetFolder);
   }
 
   // Copy
   if (fs.lstatSync(source).isDirectory()) {
-    files = fs.readdirSync(source);
-    files.forEach((file) => {
+    const files = fs.readdirSync(source);
+
+    for (const file of files) {
+      // Build source name
       const curSource = path.join(source, file);
+
+      // Call this function recursively as long as source is a directory
       if (fs.lstatSync(curSource).isDirectory()) {
         copyFolderRecursiveSync(curSource, targetFolder);
       } else {
-        const curTarget = path.join(targetFolder, path.basename(curSource));
-        // Check if the target file already exists and skip the copy in such a case to
-        // 1. avoid an exception that the file cannot be written if it is already opened
-        //    by a running instance of the pkg-packaged app
-        // 2. speed up the second, third, etc. app start if (in best case) all files
-        //    are already existing
+        // Current source is a file, it must be available on the real filesystem
+        // instead of the virtual snapshot file system to load it by process.dlopen.
+        //
+        // Before we try to copy we do some checks.
         // See https://github.com/vercel/pkg/issues/1589 for more details.
-        if (!fs.existsSync(curTarget)) {
-          fs.copyFileSync(curSource, curTarget);
+
+        // Build target file name
+        const curTarget = path.join(targetFolder, path.basename(curSource));
+
+        if (fs.existsSync(curTarget)) {
+          // Target file already exists, read source and target file...
+          const curSourceContent = fs.readFileSync(curSource, {
+            encoding: 'binary',
+          });
+          const curTargetContent = fs.readFileSync(curTarget, {
+            encoding: 'binary',
+          });
+
+          // ...and calculate checksum from source and target file
+          const curSourceHash = createHash('sha256')
+            .update(curSourceContent)
+            .digest('hex');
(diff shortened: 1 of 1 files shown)


prelude/bootstrap.js
```diff
@@ -197,10 +197,10 @@ function copyFolderRecursiveSync(source, target) {
...
         copyFolderRecursiveSync(curSource, targetFolder);
       } else {
-        fs.copyFileSync(
-          curSource,
-          path.join(targetFolder, path.basename(curSource))
-        );
+        const curTarget = path.join(targetFolder, path.basename(curSource));
+        if (!fs.existsSync(curTarget)) {
+          fs.copyFileSync(curSource, curTarget);
+        }
```
robertsLando, line 203: The change is good, I would add a comment here explaining why we need this (maybe also add a reference to the issue solved)

robertsLando (changes requested):

prelude/bootstrap.js
renkei: Sure, I will add this

renkei: Done, [0269890](https://github.com/vercel/pkg/pull/1611/commits/02698905071de369f1a9a235eb142e0796387200)

robertsLando (approved):

jesec:
Hmm. In some cases, we may want to overwrite the file:

1. File extracted becomes corrupted, caused by forced termination of the executable, a power loss or other reasons. 
2. File extracted becomes outdated. 

I think we should implement an `isSameFile` check so we overwrite it if necessary, or just `try catch do nothing` the statement so we overwrite the file if possible.

renkei:
> 1. File extracted becomes corrupted, caused by forced termination of the executable, a power loss or other reasons.

Agree, good point.

> 2. File extracted becomes outdated.

Hmm, wouldn't a updated pkg-packaged app with another `.node` file result in a new hash folder in tmp? Anyway, with the `isSameFile` check this should also be covered.
 
For security reason, I would prefer to fail / stop the app start if the `isSameFile` check results in false and the copy action fails for some reason (not sufficient acces rights, file already opened by another process, etc.). In other words: Once we have detected that the `.node` file is not the expected one and we are not able to overwrite it we should not trust the existing one.

What do you think? I can update the PR, just let me know if e.g. a SHA-256 check is sufficient for you for the `isSameFile` check or if you prefer another way to implement such a method.

luckyyyyy:
My application will start multiple workers, they will read the .node file at the same time, and if it does not exist, will they write to disk at the same time?
A lot of questions will happen if we need to consider this scenario.

jesec:
> > I think we should implement an isSameFile check so we overwrite it if necessary, or just try catch do nothing the statement so we overwrite the file if possible.
> 
> We already calculate the sha of the file, isn't that enough?

We calculated the hash of the file to be dlopen-ed, but not the hashes of all contents under the `modulePkgFolder` directory. 

Additionally, the hash only determines the folder name in the temporary directory, not if the file on the disk is the same as the file in the snapshot.

robertsLando:
> I think we should implement an isSameFile check so we overwrite it if necessary, or just try catch do nothing the statement so we overwrite the file if possible.

Ok so @renkei I think you could try with the second approach for now, it's easier

renkei:
Which method you would prefer for a `isSameFile` check? A SHA-256 check? Will this slow down the app start or is this irrelavant regarding start time if we will do this on each start for each file?

AlejandroSotoCastro (approved):
This fixed a mayor bug for us

robertsLando:
> Which method you would prefer for a isSameFile check? A SHA-256 check? Will this slow down the app start or is this irrelevant regarding start time if we will do this on each start for each file?

jesec:
> Which method you would prefer for a `isSameFile` check? A SHA-256 check? Will this slow down the app start or is this irrelevant regarding start time if we will do this on each start for each file?

Correctness is more essential than performance. Let's ensure correctness first, then try to optimize it. 

Actually, with either method (`isSameFile` or `try catch`), there is a risk of race condition when multiple same processes are launched. We may want to have a temporary directory for every process instead, but the cleanup and disk usage would be an issue.

luckyyyyy:
>  there is a risk of race condition when multiple same processes are launched.

@jesec This is what I worry about. My application will start multiple worker_threads and use a lot of C/Rust modules written by myself. I temporarily move all node files to the main process and require it, which is very helpless

luckyyyyy:
@jesec Is it possible to ask the user to tag node modules to load them early in a single process?

robertsLando:
What about to use process pid to identify the processe that is using a specific module and store this info somewhere in the temporary directory ?

renkei:
I've update the PR to also perform the copy even if the file already exists but is outdated.

There is still no solution for the use case that multiple instances of the app might start simultaneously. I see only two options:

1. The start of instance 2, 3, 4, etc. is delayed, so that instance 1 can finish the copy
2. Each instance don't care about other instances because each instance has its own directory

Solution 1 must be implemented by the user that tries to start multiple instances because we cannot implement a delayed check&copy (with some retries to give the other instance time to finish the copy) in bootstrap.js because everything inside process.dlopen must be synchronous. To be honest, I would prefer this solution, it is much easier for the user to control the start of multiple instances than for pkg.

Solution 2 sounds interesting but has so much disadvantages. It makes things more complicated and can produce a lot of spam on disk since we copy the same files again and again, they cannot be shared across multiple instances anymore. I'm not sure if it's worth the effort, maybe we're leaving the KISS pattern too much here...

luckyyyyy:
I loaded all node files in the main process.

image
image

robertsLando (approved):

jesec (approved):
It would be more correct and less error-prone to let each process have its own temporary directory. I have implemented the approach: https://github.com/jesec/pkg/commit/71f697d3ec6c20e4e91ffe30c95b91e953ca6674, but I am still sorting out issues with Node 12 on Windows at the moment. Duplication would be annoying but correctness is more essential. 

Nonetheless, this PR does provide some improvements over the status quo and we can get it merged first.

renkei:
> I have implemented the approach: https://github.com/jesec/pkg/commit/71f697d3ec6c20e4e91ffe30c95b91e953ca6674

Nice, I will try this out. Thanks!

> Duplication would be annoying but correctness is more essential.

Beside the duplication (which is not really an issue, I guess, as long as everything is removed before the app exits) this has another effect: Native modules cannot be shared across instances, no matter whether multiple instances are running simultaneously or not because the temporary folder name depends now on the process ID and `randomFillSync()` which makes the name practically unique. This avoids write conflicts but results in copy actions from virtual to real file system whenever an instance starts. Therefore the instance will start slower, which could be irrelevant or clearly noticeable. It depends on the number and size of native modules that are copied. But this is probably the price for correctness. I don't know if this is really an issue, I just want to mention this.

luckyyyyy:
> It would be more correct and less error-prone to let each process have its own temporary directory. I have implemented the approach: [jesec@71f697d](https://github.com/jesec/pkg/commit/71f697d3ec6c20e4e91ffe30c95b91e953ca6674), but I am still sorting out issues with Node 12 on Windows at the moment. Duplication would be annoying but correctness is more essential.
> 
> Nonetheless, this PR does provide some improvements over the status quo and we can get it merged first.

Verified, has run successfully, but affects startup speed. But there is no other better solution

PR #1645, opened 2022-05-30
feat: allow to specify custom babel plugins

robertsLando:
This can easily be done using `PKG_BABEL_PLUGINS` env var

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,11 +268,12 @@ The startup time of the application might be reduced slightly.
 
 ### Environment
 
-| Var            | Description                                                                               |
-| -------------- | ----------------------------------------------------------------------------------------- |
-| PKG_CACHE_PATH | Used to specify a custom path for node binaries cache folder. Default is `~/.pkg-cache`   |
-| PKG_IGNORE_TAG | Allows to ignore additional folder created on `PKG_CACHE_PATH` matching pkg-fetch version |
-| MAKE_JOB_COUNT | Allow configuring number of processes used for compiling                                  |
+| Var               | Description                                                                               |
+| ----------------- | ----------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


jesec:
Hmmm... We definitely don't want to have more env variables. 

Additionally, any Babel plugins have to be in the dependency (node_modules) tree of `pkg`. It is improbable to add them all as dependencies. 

Users should not run pkg on an entrypoint that's otherwise not runnable on Node.js runtime without transpilation. Users should run the transpilation first and then run pkg.

Thus, additional Babel plugins are definitely out-of-scope for pkg, and I don't think we can move forward in any way with this idea.

robertsLando:
@jesec In my case I needed to add `'classPrivateMethods'` and it doesn't require a plugin to be installed. It is used by a module of my application. Should I add that to babel plugins by hardcoding it? 

BTW I don't see any issue in adding env vars, they just add advanced functionalities to the application, them are not intended to be used by all users

jesec:
> @jesec In my case I needed to add `'classPrivateMethods'` and it doesn't require a plugin to be installed. It is used by a module of my application. Should I add that to babel plugins by hardcoding it here so?
> 
> BTW I don't see any issue in adding env vars, they just add advanced functionalities to the application, them are not intended to be used by all users

See https://github.com/vercel/pkg/pull/1249#issuecomment-1111543386.

Unless we want to diverge from this, we can't have additional Babel plugins. 

I am open to discussion about this, though. It might make some sense to shift the burden of the correctness check to users. In that case, we can use `preset/env` with `targets.node` set to the latest Node supported by us.

robertsLando:
> we can use preset/env with targets.node set to the latest Node supported by us.

How can we do this so?

Issue #1356, opened 2021-11-03
Cant run executable from cmd spawned by itself


ErAz7:
## Problem
PKG throws erroor when running windows executable from a process spawned by itself

## Steps to Reproduce
lets say I have a simple nodejs app like this:
```js
import cp from 'child_process';
cp.exec('start cmd');
```
so obviously it starts a new cmd every time it gets executed. Now I build it with PKG into `start.exe`, then I open a cmd in the folder and run `start.exe`. It works fine and opens a new cmd, now, if I run `start.exe` in this new cmd, I'll get the following error:
```
pkg/prelude/bootstrap.js:1523
      if (error.code !== 'MODULE_NOT_FOUND') throw error;
                                             ^

TypeError: Cannot read property 'charAt' of undefined
[90m    at Function.Module._resolveLookupPaths (internal/modules/cjs/loader.js:617:15)[39m
[90m    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:834:20)[39m
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1521:44)
[90m    at Function.Module._load (internal/modules/cjs/loader.js:725:27)[39m
    at Function.runMain (pkg/prelude/bootstrap.js:1558:12)
[90m    at internal/main/run_main_module.js:17:47[39m
```

Apparently, it cannot find the file name because `argument[0]` is `undefined` while it's supposed to be a path

CC @leerob @jesec @igorklopov

ErAz7:
btw this works fine when I directly run the source with node

SamuelCharpentier:
I'm having the same issue!

goshander:
same issue, run itself, even with bash sub process

CheneyWong:
same

orgrimarr:
PKG modify the node native child_process lib. (prelude\bootstrap.js line **1990**).

It add a var env `PKG_EXECPATH`  wich is used in (prelude\bootstrap.js line **77**).

```js
if (process.env.PKG_EXECPATH === EXECPATH) {
  process.argv.splice(1, 1); // Remove PKG_DUMMY_ENTRYPOINT

  if (process.argv[1] && process.argv[1] !== '-') {
    // https://github.com/nodejs/node/blob/1a96d83a223ff9f05f7d942fb84440d323f7b596/lib/internal/bootstrap/node.js#L269
    process.argv[1] = path.resolve(process.argv[1]);
  }
} else {
[...]

PR #1647, opened 2022-05-31
fix: `spawn` not working

orgrimarr:
Fixes https://github.com/vercel/pkg/issues/1356


robertsLando (commented):
Instead of using an env var I suggest to auto-detect the spawn of itself in `childProcess.spawn` / `childProcess.spawnSync` calls, it makes more sense. Also if you want to test locally just run the single test you want not all the others. You pushed a huge pnpm-loc.yaml file, remove it

robertsLando (changes requested):
Instead of using an env var I suggest to auto-detect the spawn of itself in `childProcess.spawn` / `childProcess.spawnSync` calls, it makes more sense. Also if you want to test locally just run the single test you want not all the others. You pushed a huge pnpm-loc.yaml file, remove it

.gitignore
```diff
@@ -20,4 +20,8 @@ examples/express/node_modules
 # Editors
 .vscode/
 yarn-error.log
-tsconfig.tsbuildinfo
\ No newline at end of file
+tsconfig.tsbuildinfo
+
+# Tests temp
+*.exe
+*.bak
```
robertsLando, line 27: remove changes to this file

robertsLando (changes requested):

robertsLando:
Also fix lint issues please

.gitignore
orgrimarr: Should be fixed now

robertsLando (changes requested):
A test is still failing

robertsLando:
Ok

lamualfa:
This is great. Please merge it.

PR #1648, opened 2022-06-01
detector: use Babel AST and default plugins

jesec:
Allow new ECMAScript features by default.

Support language features supported by the latest version of Node.js supported, instead of the oldest. 

pkg only uses Babel to parse dependencies. It does NOT transform users' sources. It is the users' responsibility to make sure that their code can run on the target Node.js version. 

https://babeljs.io/docs/en/babel-parser#latest-ecmascript-features

Refs: https://github.com/vercel/pkg/pull/1249#issuecomment-1111543386
Closes: #1249, #1645

### DIFF
--- a/lib/detector.ts
+++ b/lib/detector.ts
@@ -1,32 +1,38 @@
-/* eslint-disable operator-linebreak */
-/* eslint-disable prefer-const */
-
-import { generate } from 'escodegen';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import * as babelTypes from '@babel/types';
 import * as babel from '@babel/parser';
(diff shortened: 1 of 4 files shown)


jesec:
@robertsLando Can you check if this PR fixes things for you?

robertsLando:
@jesec I will be on holiday until Monday, if you want to give it a try in the meanwhile just create e file that imports aedes-persistence-mongodb package and try to pack it :) 

Thanks for this pr 🙏🏻

robertsLando (approved):
Tested and it seems working 🎉 LGTM thanks!

Issue #1695, opened 2022-07-13
Bug in target version for napi


rightaway:
### What version of pkg are you using?

5.7.0

### What version of Node.js are you using?

16.15.0

### What operating system are you using?

linux

### What CPU architecture are you using?

amd64

### What Node versions, OSs and CPU architectures are you building for?

node16-linux-64

### Describe the Bug

I get these warnings when I run pkg with sharp as a dependency.

```
prebuild-install WARN This package does not support N-API version v16.15.0
prebuild-install WARN install No prebuilt binaries found (target=v16.15.0 runtime=napi arch=x64 libc= platform=linux)
```

@lovell the maintainer of sharp says it's a bug in pkg in https://github.com/vercel/pkg/blob/62f6da937a915dde824e30e5b9af7f707ba391fb/lib/producer.ts#L228-L229 because it is hard-coding the target to be the version of Node.js, which is incorrect when the runtime is `napi`. https://github.com/lovell/sharp/issues/3289#issuecomment-1181823295

### To Reproduce

Run pkg with sharp as a dependency. With latest version 0.30.7 there are warnings but not with older version 0.27.2.

PR #1699, opened 2022-07-19
producer: properly call "prebuild-install" if N-API is used

jesec:
Don't pass "--target" if the package uses N-API.

Bug: #1695

### DIFF
--- a/lib/producer.ts
+++ b/lib/producer.ts
@@ -221,19 +221,24 @@ function nativePrebuildInstall(target: Target, nodeFile: string) {
     fs.copyFileSync(nodeFile, `${nodeFile}.bak`);
   }
 
+  const napiVersions = JSON.parse(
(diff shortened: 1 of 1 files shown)


lib/producer.ts
```diff
@@ -221,19 +221,24 @@ function nativePrebuildInstall(target: Target, nodeFile: string) {
...
+  )?.binary?.napi_versions;
+
+  const options = [
+    '--platform',
+    platform[target.platform],
+    '--arch',
+    target.arch,
+  ];
+
+  if (napiVersions == null) {
```
robertsLando, line 235: Shouldn't you change `--runtime` option based on this? https://github.com/prebuild/prebuild-install#help

robertsLando (changes requested):

lib/producer.ts
jesec: See #1695. prebuild-install already does that.

robertsLando (approved):

PR #1745, opened 2022-09-05
Add patch for ffi-napi and ref-napi

### LINKED ISSUE #1744: ffi-napi unable to load embedded library
### What version of pkg are you using?

5.8.0

### What version of Node.js are you using?

16.17.0

### What operating system are you using?

Ubuntu 20.04

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

node16-linux-x64

### Describe the Bug

I'm being unable to run a pkg'ed application that loads a `.so` file on runtime. This issue seemed at first possibly related to https://github.com/vercel/pkg/issues/1349, but in the end I think it is a different thing.

The system is Linux Mint 20.3 === Ubuntu 20.04

An application that uses [vosk](https://www.npmjs.com/package/vosk) installs a shared library under `node_modules/vosk/lib/linux-x86_64/libvosk.so`; it also uses `ffi-napi` and `ref-napi` to load it, but the pkg'ed executable fails running with the following error:

```sh
$ dist/bin/myapp
pkg/prelude/bootstrap.js:1740
      throw error;
      ^

Error: Dynamic Linking Error: /snapshot/home/code/myapp/node_modules/vosk/lib/linux-x86_64/libvosk.so: cannot open shared object file: No such file or directory
    at new DynamicLibrary (/snapshot/home/code/myapp/node_modules/ffi-napi/lib/dynamic_library.js:75:11)
    at Object.Library (/snapshot/home/code/myapp/node_modules/ffi-napi/lib/library.js:47:10)
    at Object.<anonymous> (/snapshot/home/code/myapp/node_modules/vosk/index.js:84:21)
    at Module._compile (pkg/prelude/bootstrap.js:1794:22)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at Module.require (pkg/prelude/bootstrap.js:1719:31)
    at require (node:internal/modules/cjs/helpers:94:18)
```

I'm running pkg with this command: `pkg pkg.config.json`, where the file contents are as follows:

```json
{
  "name": "myapp",
  "bin": "dist/main.js",
  "private": true,
  "pkg": {
    "assets": [
      "node_modules/ffi-napi/build/Release/ffi_bindings.node",
      "node_modules/ref-napi/prebuilds/linux-x64/node.napi.node",
      "node_modules/vosk/lib/linux-x86_64/libvosk.so"
    ],
    "outputPath": "dist/bin/",
    "targets": ["node16-linux-x64"]
  }
}
```

Lastly, these are the contents of `/tmp/pkg/`:

```sh
$ tree -a /tmp/pkg/
/tmp/pkg/
├── 94544448971968a49fabf8226f94e981d9a51f06abcecd8a98dd023986f814c2
│   └── ffi-napi
│       ├── build
│       │   └── Release
│       │       └── ffi_bindings.node
│       ├── lib
│       │   ├── bindings.js
│       │   ├── callback.js
│       │   ├── cif.js
│       │   ├── cif_var.js
│       │   ├── dynamic_library.js
│       │   ├── errno.js
│       │   ├── ffi.js
│       │   ├── _foreign_function.js
│       │   ├── foreign_function.js
│       │   ├── foreign_function_var.js
│       │   ├── function.js
│       │   ├── library.js
│       │   └── type.js
│       └── package.json
└── c0242e741053036807a4c1e2c3ed290cd56d19e97f5e106570434ed84191500c
    └── ref-napi
        ├── lib
        │   └── ref.js
        ├── package.json
        └── prebuilds
            └── linux-x64
                └── node.napi.node
```

My suspicion is that, given how pkg handles `.node` modules ([Native addons](https://github.com/vercel/pkg#native-addons)), the `ffi-napi` and `ref-napi` modules get extracted to the host filesystem at `/tmp/pkg/`, but then these cannot find the `libvosk.so` file that only exists within the `dist/bin/myapp` executable.

Do you think I'm right there? Is there any way to help `ffi-napi` finding the file it is expecting, to make the application work?

### Expected Behavior

I'd expect that the `libvosk.so` file is found and successfully loaded by the FFI library.

### To Reproduce

**package.json**
```json
{
  "name": "myapp",
  "version": "0.0.1",
  "dependencies": {
    "vosk": "0.3.39"
  },
  "devDependencies": {
    "pkg": "5.8.0"
  }
}
```

**pkg.config.json**
```json
{
  "name": "myapp",
  "bin": "main.js",
  "private": true,
  "pkg": {
    "assets": [
      "node_modules/ffi-napi/build/Release/ffi_bindings.node",
      "node_modules/ref-napi/prebuilds/linux-x64/node.napi.node",
      "node_modules/vosk/lib/linux-x86_64/libvosk.so"
    ],
    "outputPath": "./",
    "targets": ["node16-linux-x64"]
  }
}
```

**main.js**
```js
const vosk = require("vosk");
```

**Commands**
```sh
$ npm install
$ npx pkg --debug pkg.config.json
$ rm -rf /tmp/pkg/
$ DEBUG_PKG=1 ./myapp
```

j1elo:
This is needed because the ffi-napi's native addon for Node is extracted by pkg to the host filesystem, thus any shared library file that is to be loaded must also be extracted.

Do not confuse the Node's badly named `process.dlopen` (which is already handled by pkg's bootstrap script, but is also useless for ffi-napi) with the actual `dlopen` function pointer that ffi-napi loads from the system's library API.

Fixes #1744

### DIFF
--- a/test/test-79-npm/ffi-napi/ffi-napi.js
+++ b/test/test-79-npm/ffi-napi/ffi-napi.js
@@ -1,0 +1,4 @@
+'use strict';
+
+const vosk = require('vosk');
+console.log('ok');
--- a/dictionary/ffi-napi.js
+++ b/dictionary/ffi-napi.js
@@ -1,0 +1,34 @@
+'use strict';
+
+module.exports = {
+  pkg: {
+    patches: {
+      // This patch reimplements the file extraction logic found in the
+      // `prelude/bootstrap.js` file's dlopen() function. This is needed
+      // because pkg's dlopen() ends up calling Node's process.dlopen(),
+      // which is not appropriate for loading shared libraries with ffi-napi.
+      'lib/dynamic_library.js': [
+        'this._path = path;',
+        `
+        if (path.startsWith('/snapshot/')) {
+          const Fs = require('fs');
+          const moduleContent = Fs.readFileSync(path);
+          const hash = require('crypto')
+            .createHash('sha256')
+            .update(moduleContent)
+            .digest('hex');
+          const Path = require('path');
+          const tmpFolder = Path.join(require('os').tmpdir(), 'pkg', hash);
+          const newPath = Path.join(tmpFolder, Path.basename(path));
+          if (!Fs.existsSync(tmpFolder)) {
+            Fs.mkdirSync(tmpFolder, { recursive: true });
+            Fs.copyFileSync(path, newPath);
+          }
+          path = newPath;
+        }
+        this._path = path;
+        `,
+      ],
+    },
+  },
+};
--- a/dictionary/ffi-napi.js
+++ b/dictionary/ffi-napi.js
@@ -2,6 +2,7 @@
 
 module.exports = {
   pkg: {
+    assets: ['build/Release/ffi_bindings.node'],
     patches: {
       // This patch reimplements the file extraction logic found in the
       // `prelude/bootstrap.js` file's dlopen() function. This is needed
--- a/dictionary/ref-napi.js
+++ b/dictionary/ref-napi.js
@@ -1,0 +1,7 @@
+'use strict';
+
+module.exports = {
+  pkg: {
+    assets: ['prebuilds/linux-x64/node.napi.node'],
+  },
+};
--- a/test/test-79-npm/ffi-napi/package.json
+++ b/test/test-79-npm/ffi-napi/package.json
@@ -1,0 +1,6 @@
+{
+  "private": true,
+  "dependencies": {
+    "vosk": "0.3.39"
+  }
+}


robertsLando (approved):
LGTM

j1elo:
Things this PR might be lacking, but I don't know internal project's "howto" in order to add adequately:

* An `assets` array? The `ffi-napi` package installs a node addon that is found in `node_modules/ffi-napi/build/Release/ffi_bindings.node`, so I'm not sure if this patch should include something like:

    ```js
    assets: [
      "build/Release/ffi_bindings.node",
    ]
    ```

* A test in `test/`
    * Would `test/test-79-npm/ffi-napi/ffi-napi.js` be the correct place?
    * What would the contents of `package.json` be? I've checked a couple at random, and they all seem to just contain the line `"private": true` but no `"dependencies"` at all.

robertsLando:
Afaik patches can also contain assets so you can add them there.

> Would test/test-79-npm/ffi-napi/ffi-napi.js be the correct place?

Yes

> What would the contents of package.json be

You can add the dependencies that you prefer in package.json. There are some tests that already do that

j1elo:
Ok so here's the thing: `ffi-napi` uses `ref-napi` as a dependency.

`ffi-napi` installs `"node_modules/ffi-napi/build/Release/ffi_bindings.node"`,
and `ref-napi` installs `"node_modules/ref-napi/prebuilds/linux-x64/node.napi.node"`.

So to make it work properly, I'm thinking that an additional patch would be needed to add `ref-napi`'s addon module to the assets.

j1elo:
> > What would the contents of package.json be
> 
> You can add the dependencies that you prefer in package.json. There are some tests that already do that

So for example, would this be a good place to load some other dependency, such like Vosk (which is what I've used for my tests) and trying to load the `libvosk.so` file in the test? That would be a good way to verify that the `ffi-napi` patch is actually working with a real-world library file.

Or should other mechanism be used for testing? Is there a shared library that can be used instead?

robertsLando:
> So for example, would this be a good place to load some other dependency, such like Vosk (which is what I've used for my tests) and trying to load the libvosk.so file in the test? That would be a good way to verify that the ffi-napi patch is actually working with a real-world library file.

Yes

j1elo:
Ok with the latest commits I've added a test for `ffi-napi`, and an additional patch for its dependency `ref-napi`, which also contains a native Node addon that should be included as asset.

test/test-79-npm/ffi-napi/ffi-napi.js
```diff
@@ -0,0 +1,4 @@
+'use strict';
+
+const vosk = require('vosk');
```
robertsLando, line 3: this test is not packaging the app, check https://github.com/vercel/pkg/tree/main/test/test-80-compression-node-opcua as an example. Remember to also add a gitignore for node_modules

robertsLando (changes requested):

test/test-79-npm/ffi-napi/ffi-napi.js
j1elo: I'm seeing the `test/test-79-npm/main.js` file and getting confused because I don't know to what extent the testing utility scripts already do stuff.

It seems like commands such as [npm install](https://github.com/vercel/pkg/blob/main/test/test-79-npm/main.js#L232), running [pkg](https://github.com/vercel/pkg/blob/dd9de59c9fca2751bf5d22b57bd9b03d43e85e80/test/test-79-npm/main.js#L303), and even [vacuum](https://github.com/vercel/pkg/blob/dd9de59c9fca2751bf5d22b57bd9b03d43e85e80/test/test-79-npm/main.js#L420)('node_modules') are already done by this main script, so I'm wondering what should actually be done by the test.

In essence this particular test should simply consist on installing `vosk`, and running a main.js file that contains `require('vosk')`. If the shared library load doesn't work, the execution would fail. But I lack the knowledge to do this within the testing framework of *pkg*, and also I'm not sure if there are multiplatform considerations that need to be addresses too.

(I'll be out of town since tomorrow, so I'm afraid I won't be able to look into this in the following weeks)

Issue #1574, opened 2022-04-01
`Cannot generate bytecode` / `pkg fails to run "codesign" utility` (even when codesign is present)


ryanblock:
### What version of pkg are you using?

5.5.2

### What version of Node.js are you using?

v14.18.2

### What operating system are you using?

macOS 12.2.1 (21D62)

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

node14-macos-x64

### Describe the Bug

Running `npx pkg` fails with the following error:
```
> pkg@5.5.2
> Fetching base Node.js binaries to PKG_CACHE_PATH
  fetched-v14.18.3-macos-x64          [====================] 100%

/Users/ryan/.pkg-cache/v3.2/fetched-v14.18.3-macos-x64-signed: is already signed
> Error! Cannot generate bytecode
  pkg fails to run "codesign" utility. Due to the mandatory signing
  requirement of macOS, executables must be signed. Please ensure the
  utility is installed and properly configured.
```

However, `codesign` is indeed installed and available:

```
⮑  which codesign
/usr/bin/codesign
⮑  codesign
Usage: codesign -s identity [-fv*] [-o flags] [-r reqs] [-i ident] path ... # sign
       codesign -v [-v*] [-R=<req string>|-R <req file path>] path|[+]pid ... # verify
       codesign -d [options] path ... # display contents
       codesign -h pid ... # display hosting paths
```

(Related: is the `fetched-v14.18.3-macos-x64-signed: is already signed` line a red herring?)

### Expected Behavior

I expect pkg to successfully run or give me a useful, usable error that I can use to fix the issue.

### To Reproduce

Steps to reproduce presented above in the bug description; of course happy to narrow down further if you have any ideas on how I can!

brianunlam:
I had the same issue. As a workaround you can go to the source code and add a "-f" flag when they call the codesign utility:


https://github.com/vercel/pkg/blob/b63df87586dfb0a5f56e8698f753846edee4a498/lib/mach-o.ts#L62

ryanblock:
@brianunlam that's a nice workaround, did the trick for me. It's super brittle for obvious reasons, @jesec @leerob any reason not to incorporate this into the `codesign` call permanently so we can close this issue?

ryota-ka:
`$ sed -e s/--sign/-s/ -i ./node_modules/pkg/lib-es5/mach-o.js` did the trick for me, FYI.

[...]

PR #1756, opened 2022-09-14
fix: add force flag to codesign to avoid already signed error

brianunlam:
It adds force flag to codesign utility to avoid errors when an existing sign prevents a new sign.
It solves https://github.com/vercel/pkg/issues/1574


robertsLando (approved):

lattwood:
Hey, anything I can do to help get this merged in?

robertsLando:
I would like to know @jesec opinion about this as he has more experience then me regarding mac OS things, dunno if that flag could break older OS?

bro4all:
The flag is just a force flag, it is very old and compatible with other os

134130:
The flag is compatible and It doesn't take any wrong effects.
> When signing, causes codesign to replace any existing signature on the path(s) given.
> Without this option, existing signatures will not be replaced, and the signing operation fails.
from `man codesign` :D

robertsLando:
@jesec ping

ryanblock:
Bump

ryanblock:
@jesec bump again

134130:
@brianunlam @robertsLando 
The auto-merge is not available since node16's prettier is failing.
The prettier is warning 'print-width' is over.

Would you resolve this issue? I don't have any permission to commit this.
Simple command: `prettier -w lib/mach-o.ts`

Mikescops:
@robertsLando can a release version be tagged with this change?

robertsLando:
@Mikescops Need to speak with @leerob and @baparham about that

PR #1757, opened 2022-09-14
chore: refactor dictionary handling

ignatiusmb:
Moves dictionary files into a single object in a file, this should reduce bundle size and slightly improve performance because we'll be directly importing the dictionary object, skipping the middleman, so to speak.

I tried to reuse some of the types, but it seems some of them are unused or not quite up to date with the current code, so I updated them as well and tried to match it with the current code as much as possible.


robertsLando (approved):

robertsLando (approved):

robertsLando:
cc @jesec

robertsLando:
@ignatiusmb WOuld like to know @jesec opinion on this before merging

ignatiusmb:
Another pair of eyes never goes wrong 😄

ignatiusmb:
What's left to do?

robertsLando (approved):

test/test-77-compare-dicts-and-tests/main.js
```diff
@@ -6,18 +6,9 @@ const fs = require('fs');
...
-    return config !== '.eslintrc.json';
-  })
-  .map(function (config) {
-    assert.strictEqual(config.slice(-3), '.js');
-    return config.slice(0, -3);
-  });
-
-let tests = fs.readdirSync('../test-79-npm');
-tests = tests.filter(function (test) {
+const configs = Object.keys(require('../../lib-es5/dictionary').default);
```
baparham, line 9: presumably this all of the sudden needs to require the `lib-es5` version because it's the transpiled javascript, not the direct javascript it was previously, right?

test/test-77-compare-dicts-and-tests/main.js
```diff
@@ -6,18 +6,9 @@ const fs = require('fs');
...
+
+const tests = fs.readdirSync('../test-79-npm').filter(function (test) {
```
baparham, line 11: There's not really an advantage here to inlining the readdir and the filter, I'd rather you extract the filter function into a named function defining what the filter is actually doing here, like `excludeIsolatorTests` or something

lib/walker.ts
```diff
@@ -670,29 +670,30 @@ class Walker {
...
         }
-      } else if (typeof patch[i] === 'string') {
+      } else if (typeof patch === 'string') {
         // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
         // function escapeRegExp
-        const esc = patch[i].replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        const esc = patch.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
         const regexp = new RegExp(esc, 'g');
-        body = body.replace(regexp, patch[i + 1]);
+        body = body.replace(regexp, patches[i + 1] as string);
```
baparham, line 696: why do you need to cast as a string here? I think that may be a symptom of an earlier typing problem that should be resolved as part of this refactor.

lib/walker.ts
```diff
@@ -670,29 +670,30 @@ class Walker {
...
+          body = patches[i + 1] + body;
+        } else if (patch.do === 'append') {
+          body += patches[i + 1];
         }
-      } else if (typeof patch[i] === 'string') {
+      } else if (typeof patch === 'string') {
         // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions
         // function escapeRegExp
-        const esc = patch[i].replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        const esc = patch.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
baparham, line 694: while you are refactoring, please go ahead and rename variables to meaningful names, like `escapedRegEx` in this case.

lib/walker.ts
baparham: I also can't see how this doesn't hit an array out of bounds exception. The current `for` loop logic is convoluted and should probably be refactored to be readable to future contributors. What is it doing? Can parts be broken out into meaningful function names?

baparham (commented):
I'm on the fence on this one, can you clarify what problem you are trying to solve? This is obviously a refactor, but there are still things I'd like to see improved while you are in here so that it's not just "different and not understandable" but "different and _more_ understandable" to future contributors.

on one hand, it is nice to have the modules in their own separate files, since they do not relate to one another. Changes in one module file should not be part of the history when looking at another module. This pulls them all into one mega-file. Granted, they don't change often, so churn should be low.

On the other hand, the dictionary parsing code got much simpler, and in theory, more maintainable, which is a good thing.

baparham:
The "not understandable" reference was to existing code and variable naming, which you (rightly) continued to follow, not a comment on your code specifically. I'm just looking for the code quality to improve more while you are touching these parts of the code.

test/test-77-compare-dicts-and-tests/main.js
ignatiusmb: I don't see any benefit from doing that and I don't think we should touch anything more than what this PR does, especially with old tests like this.

ignatiusmb:
> I'm just looking for the code quality to improve more while you are touching these parts of the code.

I would love to do that, but the changes here are already huge as it is, I would prefer to have that in a separate PR as it may get out of scope if we start touching a lot of unrelated code here.

lib/walker.ts
ignatiusmb: I reapproached this refactor from the types instead, it should clarify the intent better now.

ignatiusmb:
> can you clarify what problem you are trying to solve? [...] on one hand, it is nice to have the modules in their own separate files, since they do not relate to one another. Changes in one module file should not be part of the history when looking at another module. This pulls them all into one mega-file. Granted, they don't change often, so churn should be low.

To keep it simple and consistent in one file and type-checked with TypeScript, make it easier to add or amend patches in the future, and reduce inefficient `path` concatenation and `fs` calls.

ignatiusmb:
Test failures seems unrelated?

lib/walker.ts
ignatiusmb: Didn't think about it on the moment but refactoring this in any way from the types or changing how the loop works will be a breaking change to the `"patches"` API.

baparham: it seems like you've made it better with your most recent changes though. I do not know what the existing community does with these dictionary entries, if they tend to patch them themselves (in which case, yes, this would break their patches) or if they are just used by pkg itself and can be considered an implementation detail.

I'd argue that these are the latter, and as an implementation detail, can't be relied upon by anyone patching them themselves at runtime to have some sort of stable API that we'd have to semver control.

Do you have any hints about what the docs say about these dictionary entries?

test/test-77-compare-dicts-and-tests/main.js
baparham: I meant if you are going to change these lines anyway, why not answer some of the questions posed by the obscure `.filter` logic...what is that filtering doing? simply breaking that filter function into a named function that actually tells the reader what it's doing, or why would help the next person come in and also continue to improve the code next time they touch this line of code.

lib/walker.ts
baparham: I can't find any hint that these dictionary entries should be messed with by users. no "do this to support library XYZ" at all.

ignatiusmb: These are the only "docs" I could find regarding the patches https://github.com/vercel/pkg/wiki/Developers#how-it-works, and that's for the developers of pkg (or more accurately, maintainers). I think you might be right that these are only used internally and by pkg itself, so I'm guessing it's okay to move forward with the recent changes?

test/test-77-compare-dicts-and-tests/main.js
ignatiusmb: I see, I've pushed some changes for this file, did I understand you correctly with those changes?

lib/walker.ts
baparham: I think this conversation can be resolved then, thanks for the discussion and thought

baparham (approved):
My initial concerns have been addressed, so I see no reason this shouldn't go in. Thanks for the patience @ignatiusmb

robertsLando (changes requested):
@ignatiusmb Could you resolve conflicts with master please?

lib/walker.ts
```diff
@@ -660,39 +656,24 @@ class Walker {
...
+      if (patch.command === 'replace') {
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping
+        const escaped = patch.from.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+        body = body.replace(new RegExp(escaped, 'g'), patch.to);
+      } else if (patch.command === 'append') {
+        body += patch.source;
+      } else if (patch.command === 'erase') {
+        body = patch.source;
+      } else if (patch.command === 'prepend') {
+        body = patch.source + body;
```
robertsLando, line 676: Wait this is a breaking change... There are users that provide dictionary from package.json, in case them were using the old replace format this would stop working, could you add a fallback to also make replace work as before?

test/test-50-package-json-3/package.json
```diff
@@ -3,8 +3,11 @@
   "pkg": {
     "patches": {
       "test-x-index.js": [
-        "require(__dirname + '/' + dataPath);",
-        "require(process.cwd() + '/' + dataPath);"
+        {
+          "command": "replace",
+          "from": "require(__dirname + '/' + dataPath);",
+          "to": "require(process.cwd() + '/' + dataPath);"
+        }
```
robertsLando, line 10: Like this case

robertsLando (changes requested):

PR #1763, opened 2022-09-20
feat: use native node fs

ignatiusmb:
There's still a test case where it requires `fs-extra`, so I'm not sure if we can completely remove the package, but at least we can move it out of the dependency tree.


robertsLando (changes requested):
I'm not sure about this changes. You have replaced lot async methods with sync ones, also fs-extra doesn't create so much overhead as it uses fs internally, it just extends it

ignatiusmb:
The rationale for replacing those async methods is because it's immediately `await`-ed and used — e.g., it's executed as `(await asyncThing()).anotherSyncMethod()`. I find that using synchronous methods for these operations a bit simpler and easier to read unless we need to do multiple things simultaneously.

> fs-extra doesn't create so much overhead as it uses fs internally, it just extends it

For this, I think it's more of a reason to not have it as a dependency, especially when we can already do the things we need with node's fs module.

Another reason is also to reduce and minimize dependencies as much as possible, recent survey shows that [managing dependencies is one of the biggest pain points when working with JS](https://2021.stateofjs.com/en-US/opinions/#js_pain_points_wins). Generally speaking, I think the less dependencies any project has, the better the experience would be, both for the maintainers and the users.

ignatiusmb:
Please let me know if you prefer to keep the async methods, I'll address them and make the changes.

robertsLando (approved):

robertsLando:
@jesec feel free to merge if it looks good to you

ignatiusmb:
What's left to do?

lib/chmod.ts
```diff
@@ -1,13 +1,13 @@
-import { chmod, stat } from 'fs-extra';
+import { promises as fs } from 'fs';
```
robertsLando, line 1:
```suggestion
import fs from 'fs/promises';
```

?

lib/walker.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable require-atomic-updates */
 
 import assert from 'assert';
-import fs from 'fs-extra';
+import { promises as fs } from 'fs';
```
robertsLando, line 4:
```suggestion
import fs from 'fs/promises';
```

robertsLando (changes requested):

ignatiusmb:
@robertsLando ES6 `fs/promises` import won't work with Node v12 (https://stackoverflow.com/a/70042193).

pcnate:
> @robertsLando ES6 `fs/promises` import won't work with Node v12 (https://stackoverflow.com/a/70042193).

NodeJS 14.x is reaching the end of its LTS support. I doubt that 12.x should still be considered.

PR #1767, opened 2022-09-21
chore: upgrade actions runners

ignatiusmb:


robertsLando (changes requested):
Could you restore `name` in steps?

ignatiusmb:
I can, though I find the uses/run commands self-explanatory, most actions usage in many repositories and even the example in the readme often omit the `name` keys, they're usually added when there's a custom `run` script with multiple lines.

@robertsLando are you willing to reconsider removing them? If not, the `actions/cache` doesn't have a `name` for it, should I also add it for consistency?

robertsLando (approved):
Ok

PR #1769, opened 2022-09-22
chore: remove unused dependencies

ignatiusmb:
See if this works, we might be able to remove a lot of dependencies that's only used (`require`-d) in tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,14 +60,6 @@
     "simple-git-hooks": ">=2.8.0",
     "typescript": "4.7.2"
   },
-  "peerDependencies": {
-    "node-notifier": ">=9.0.1"
-  },
-  "peerDependenciesMeta": {
-    "node-notifier": {
-      "optional": true
-    }
-  },
   "scripts": {
     "clean": "rimraf lib-es5",
     "build": "npm run clean && tsc",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
   "devDependencies": {
     "@babel/core": "7.18.2",
     "@types/babel__generator": "7.6.4",
-    "@types/escodegen": "0.0.7",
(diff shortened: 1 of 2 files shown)


ignatiusmb:
Seems like it's working, pushing one more change and this should be ready

jesec (approved):

package.json
```diff
@@ -60,14 +59,6 @@
     "simple-git-hooks": ">=2.8.0",
     "typescript": "4.7.2"
   },
-  "peerDependencies": {
-    "node-notifier": ">=9.0.1"
```
jesec, line 64: We do have dependency on "node-notifier" in "test-79-npm" at the moment.

jesec (dismissed):

package.json
ignatiusmb: Yes, that's what I was trying to see in the first commit, and all the tests seems to pass.

It confirms my suspicion with the `"peerDependenciesMeta"` that this package won't be installed unless specified explicitly, so it's not installed even before this change, which we can confirm by looking at the lockfile not having `"node-notifier"` anywhere.

Another thing is that "test-79-npm" seems to download all the packages by itself using `https` module, which might not be the best approach to testing these kinds of stuff, but that would also mean we may not need any of those packages in the project's `package.json`.

ignatiusmb:
What's left to do?

robertsLando (approved):

baparham:
I think it's reasonable to dismiss Jesse's review since he's no longer actively involved, and his original concern does seem to be addressed to me. I'll go ahead and dismiss the 'changes requested' flag, and document my justification here.

baparham:
@ignatiusmb perhaps a rebase and rerun of the tests and we'll see how that goes

PR #1850, opened 2023-01-21
Added work with module-alias

sacredneobi:


robertsLando (changes requested):
Please better explain what this is fixing and add a test for it

PR #1862, opened 2023-02-06
feat: support node v19

phated:
In combination with https://github.com/vercel/pkg-fetch/pull/274, this adds support for node 19.6

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -1879,7 +1879,11 @@ function payloadFileSync(pointer) {
     im = require('internal/module');
     makeRequireFunction = im.makeRequireFunction;
   } else {
-    im = require('internal/modules/cjs/helpers');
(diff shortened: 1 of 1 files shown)


robertsLando (approved):

robertsLando:
cc @baparham

phated:
@robertsLando I think `pkg-fetch` needs a release before I can update this. Is that correct? I'd love to get this released, since node 20 is less than 2 months away and we should be tracking the changes necessary through the unstable version

robertsLando:
We are working to try setting up back ARM GitHub runners, ATM a new release would miss all ARM e executables otherwise

baparham:
Perhaps this needs a rebase also @phated and then I think we can merge it in once it's updated

phated:
Rebased with the github UI

Issue #1863, opened 2023-02-06
Tesseract.js v4 - Cannot find module


some1chan:
### What version of pkg are you using?

5.8.0

### What version of Node.js are you using?

18.13.0

### What operating system are you using?

Windows

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

node18-win-x64

### Describe the Bug

When trying to use Tesseract (ex. `Tesseract.recognize()` or `Tesseract.createWorker()` the calls fail when compiled into an executable, and throws this error: 
```js
node:internal/event_target:915
  process.nextTick(() => { throw err; });
                           ^
Error: Cannot find module 'C:\snapshot\tesseract.js\node_modules\tesseract.js\src\worker-script\node\index.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:946:15)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1951:46)
    at Module._load (node:internal/modules/cjs/loader:787:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at MessagePort.<anonymous> (node:internal/main/worker_thread:194:24)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:645:20)
    at exports.emitMessage (node:internal/per_context/messageport:23:28)
Emitted 'error' event on Worker instance at:
    at [kOnErrorMessage] (node:internal/worker:290:10)
    at [kOnMessage] (node:internal/worker:301:37)
    at MessagePort.<anonymous> (node:internal/worker:202:57)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:645:20)
    at exports.emitMessage (node:internal/per_context/messageport:23:28) {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

### Expected Behavior

To be able to call functions from Tesseract.js and not error out

### To Reproduce

Run the following commands:
```powershell
# Clone the reproduction repository, and install deps
git clone https://github.com/some1chan/pkg-tesseract.js-repro.git
cd pkg-tesseract.js-repro
npm i

npm start
# Outputs: 
# > Attempting to create worker
# > Successfully created worker!

npm run build
.\dist\app.exe
# Outputs:
# > Attempting to create worker
# > 
# > node:internal/event_target:915
# >  process.nextTick(() => { throw err; });
# > ...
```

PR #1864, opened 2023-02-06
test: update tesseract.js test for v4

some1chan:
Test will still fail due to #1863, but at least it _should_ be correct if/when it's fixed.

### DIFF
--- a/dictionary/tesseract.js.js
+++ b/dictionary/tesseract.js.js
@@ -2,6 +2,6 @@
 
 module.exports = {
   pkg: {
-    scripts: ['src/node/worker.js'],
+    scripts: ['src/worker-script/node/index.js'],
   },
 };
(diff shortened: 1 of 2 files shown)


robertsLando:
In order to fix the original issue you could just change the patch here to include the correct file to scripts: https://github.com/vercel/pkg/blob/main/dictionary/tesseract.js.js#L5

The file should be instead: `src\worker-script\node\index.js`

some1chan:
Thank you so much for your help! Seems to be working now.

robertsLando (approved):

robertsLando:
@leerob Kindly merge once tests are ok 🙏🏼

Mikescops:
Can we rebase and merge this?

some1chan:
Should be rebased, let me know if there's anything I need to do on my end.

PR #1878, opened 2023-03-07
feat: add option to skip signature on macos

Mikescops:
There are scenarios where one would want to sign manually or later the binaries.

I'm adding a simple option that will skip the final codesign on macos.

### DIFF
--- a/test/test-50-reproducible/main.js
+++ b/test/test-50-reproducible/main.js
@@ -20,6 +20,7 @@ const output2 = './test-output-2.exe';
 
 utils.pkg.sync([
   '--public',
+  '--no-signature', // the signature will make the build not reproducible
   '--no-bytecode',
   '--target',
   target,
@@ -30,6 +31,7 @@ utils.pkg.sync([
 
 utils.pkg.sync([
   '--public',
+  '--no-signature',
   '--no-bytecode',
   '--target',
   target,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false # prevent test to stop if one fails
       matrix:
         node-version: [12.x, 14.x, 16.x, 18.x]
-        os: [ubuntu-latest, windows-latest] # Skip macos-latest
+        os: [ubuntu-latest, windows-latest, macos-latest]
 
     runs-on: ${{ matrix.os }}
     steps:
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ pkg [options] <input>
     --public-packages    force specified packages to be considered public
     --no-bytecode        skip bytecode generation and include source files as plain js
     --no-native-build    skip native addons build
+    --no-signature       skip signature of the final executable on macos
     --no-dict            comma-separated list of packages names to ignore dictionaries. Use --no-dict * to disable all dictionaries
     -C, --compress       [default=None] compression algorithm = Brotli or GZip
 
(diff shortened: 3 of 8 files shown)


robertsLando (approved):

Mikescops:
@baparham whenever you have time to take a look here 🙌

baparham:
Can you come up with a test for this? There seems to be one other signature test but I'm not entirely sure how you could replicate it to test this feature and only run on mac, but I welcome your thoughts, otherwise this is untested code (even though it is simple enough to 'test' in your brain)

https://github.com/vercel/pkg/blob/main/test/test-50-signature/main.js

Mikescops:
@baparham I come up with a solution that works well on my mac, but it seems like there are no CI for macos-latest currently

baparham:
Ah, of course. We should fix that, are you comfortable adding to the workflow matrix in this pr?

baparham:
let's see! Thanks

Mikescops:
Macos is damn too smart... 
image

Mikescops:
@baparham ok I made some extra fixes and I could run all tests on my computer

test/test-50-reproducible/main.js
```diff
@@ -20,6 +20,7 @@ const output2 = './test-output-2.exe';
 
 utils.pkg.sync([
   '--public',
+  '--no-signature', // the signature will make the build not reproducible
```
baparham, line 23: how did this work before? is this just applicable to running test-50-reproducible on macos now?

Mikescops:
Sounds good ✅

Mikescops:
I'd recommend dropping node12 support to lighten a bit the CI

baparham:
> I'd recommend dropping node12 support to lighten a bit the CI

I 100% agree with this. I see no reason to support an EOL node js publicly with pkg.

I'm not sure what SLA we claim in pkg though, so this probably needs some more discussion from @robertsLando and maybe @leerob

Mikescops:
I have opened a separate issue as we can move forward on this specific PR first https://github.com/vercel/pkg/issues/1886

Is it ok for you to merge?

baparham (approved):
lgtm

Mikescops:
why this merge commit?

baparham:
> why this merge commit?

It seems the repo is set up to require the branch be up to date before I can merge it. I don't really know why, but now it says it can be merged in, which I'll do in a squash commit. Perhaps the requirement is that the checks pass on the final merged result or something against `main` 🤷

Mikescops:
Weird, I made a rebase on main before pushing 🤔

Thanks for the help here!

Mikescops:
@baparham @robertsLando can we tag a release? i'd need to use the new option soon 👍

robertsLando:
@Mikescops Unfortunately this doesn't depends by us as we both don't have the permissions to make new releases. @baparham already sent a private message to @leerob for this, I think it will be ready for the end of this week.

BTW in the meanwhile you could simply `npm install vercel/pkg#master`

Mikescops:
Ok thanks, I'll wait then. I can't really do this in a production software unfortunately 😂

robertsLando:
@Mikescops Giving that you just need to use it to package the application (it's a dev dependency) that will give the same result as after the release so not really a big difference

Mikescops:
Not really, because it will be used in a brew Formula to package the app, and it is versioned, and I can't guarantee what will be added to the main branch here 'till i can update the brew Formula.

Issue #1886, opened 2023-03-17
Drop support of Node 12 as it is EOL


Mikescops:
### What version of pkg are you using?

5.8.1

### What version of Node.js are you using?

18

### What operating system are you using?

all

### What CPU architecture are you using?

all

### What Node versions, OSs and CPU architectures are you building for?

all

### Describe the Bug

Drop support of Node 12 as it is EOL.

pkg is currently supporting Node 12 in the CI tests, as this version of node is EOL we could remove it.

Side note: Node 14 will be EOL on 30 Apr 2023.

### Expected Behavior

N/A

### To Reproduce

N/A

baparham:
@leerob @robertsLando I'm all for this, I think we could probably just make the call and bump major versions to `6.0.0` and only support 16, 18, and 19 with it.

14 is EOL in about a month, so why keep it around if we are dropping old support in `6.0.0` unless we want to match the node.js timeline exactly (https://github.com/nodejs/release#release-schedule)

baparham:
the last few node 14 fixes maybe can be patch released into the 5.x stream if they release in the next month.

robertsLando:
I would keep node 14 for a while, AFAIK when we discussed about this we planned to keep LTS and latest EOL release (in this case node 14 until 16 reaches EOL)

robertsLando:
> BUT the version has been EOL by that time, so why support it if node.js is not supporting it???

I agree with you but last time we discussed this with the team the decision was that, maybe just to give a little time to users to move on, also because if we remove it it means users need to stick with a specific pkg version otherwise they cannot use an old nodejs version anymore (whatever is the reason they want to do so)

PR #1889, opened 2023-03-20
chore: remove eol nodejs in tests

Mikescops:
As per #1886, here is a PR to drop support for 12 which is EOL and 14 which is EOL in a month.

As some users may rely on those versions, this should be included in a major release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,9 +78,9 @@
     "fix": "npm run lint:style -- -w && npm run lint:code -- --fix",
     "prepare": "npm run build",
     "prepublishOnly": "npm run lint",
-    "test": "npm run build && npm run test:14 && npm run test:12 && npm run test:10 && npm run test:host",
-    "test:10": "node test/test.js node10 no-npm",
-    "test:12": "node test/test.js node12 no-npm",
+    "test": "npm run build && npm run test:18 && npm run test:16 && npm run test:14 && npm run test:host",
+    "test:18": "node test/test.js node18 no-npm",
+    "test:16": "node test/test.js node16 no-npm",
     "test:14": "node test/test.js node14 no-npm",
     "test:host": "node test/test.js host only-npm"
   },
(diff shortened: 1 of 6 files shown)


Mikescops:
The repo owner should remove the v14 requirement for the CI so this PR can be accepted.

robertsLando (changes requested):
AFAIK we planned to also keep last LTS and one EOL so I would keep the nodejs 14 for now

baparham:
I agree with Daniel, at least for the next month, we do need to keep node 14 around until it's officially EOL. What we do after that is still up for debate I suppose, but we should be able to land this and at least drop support for 12 with a major version bump.

package.json
```diff
@@ -78,10 +78,7 @@
     "fix": "npm run lint:style -- -w && npm run lint:code -- --fix",
     "prepare": "npm run build",
     "prepublishOnly": "npm run lint",
-    "test": "npm run build && npm run test:14 && npm run test:12 && npm run test:10 && npm run test:host",
-    "test:10": "node test/test.js node10 no-npm",
-    "test:12": "node test/test.js node12 no-npm",
-    "test:14": "node test/test.js node14 no-npm",
+    "test": "npm run build && npm run test:host",
```
baparham, line 81: should this be updated instead to run tests against 16, 18 and 19? I'm not exactly sure what is the goal of this existing infrastructure to test on a matrix of node versions in the runner, and run test scripts targeting specific node versions...cross-target compatibility maybe?

Mikescops: ah probably, i will limit it to the latest nodejs to test older versions
it will speed up things i guess

baparham:
@robertsLando it looks like @Mikescops fixed the requested change and put node 14 back in, but this still has your requested changes flag on it

robertsLando (approved):
Because I didn't receive any request for new review :) BTW LGTM

PR #1890, opened 2023-03-20
chore: make corruption test pass on macos

Mikescops:
In my last PR #1878, I left a test scenario as todo concerning damaging the binary. 

In fact, damaging when the code is signed had no effect as we were likely damaging the file in a non sensitive place.
To circumvent that I decided to skip the signature on macos (as other systems don't sign anyway), damage the file, then sign it manually and finally execute the test.

### DIFF
--- a/test/test-50-corrupt-executable/main.js
+++ b/test/test-50-corrupt-executable/main.js
@@ -10,11 +10,6 @@ const utils = require('../utils.js');
 assert(!module.parent);
 assert(__dirname === process.cwd());
 
-// TODO : understand why the damage is not impacting macos build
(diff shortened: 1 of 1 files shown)


robertsLando (approved):

Mikescops:
Outch, the auto-merge didn't wait for the macos test to pass...

robertsLando:
@Mikescops I'm sorry for that, seems them are not set as required tests in order to pass, let's see if they end successfully

Mikescops:
@robertsLando it went well thankfully

robertsLando:
Perfect, thanks!

PR #1913, opened 2023-04-19
Treat other node extensions (.mjs, .cjs) as js

rotu:


robertsLando (changes requested):
Not sure if it's correct to also include `.mjs` files, them could not be parsed correctly. Did you try to see if it break something? Could you add a test for this?

PR #1613, opened 2022-05-03
fix broken tests on node 12; latest pnpm requires node >= 14.19

kldzj:
Tests on node 12 are failing because pnpm requires `node >= 14.19`.

PR #1914, opened 2023-04-20
Upgrade pkg-fetch to 3.5.2

dav-is:
To use the newest versions of Node.js, we need to upgrade `pkg-fetch` ([GitHub Release](https://github.com/vercel/pkg-fetch/releases/tag/v3.5), [npm](https://www.npmjs.com/package/pkg-fetch))

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "is-core-module": "2.9.0",
     "minimist": "^1.2.6",
     "multistream": "^4.1.0",
-    "pkg-fetch": "3.4.2",
+    "pkg-fetch": "3.5.2",
     "prebuild-install": "7.1.1",
     "resolve": "^1.22.0",
     "stream-meter": "^1.0.4"
--- a/yarn.lock
+++ b/yarn.lock
@@ -2083,10 +2083,10 @@ picomatch@^2.2.3:
   resolved "https://registry.yarnpkg.com/picomatch/-/picomatch-2.3.0.tgz#f1f061de8f6a4bf022892e2d128234fb98302972"
   integrity sha512-lY1Q/PiJGC2zOv/z391WOTD+Z02bCgsFfvxoXXf6h7kv9o+WmsmzYqrAwY63sNgOxE4xEdq0WyUnXfKeBrSvYw==
 
-pkg-fetch@3.4.2:
-  version "3.4.2"
-  resolved "https://registry.yarnpkg.com/pkg-fetch/-/pkg-fetch-3.4.2.tgz#6f68ebc54842b73f8c0808959a9df3739dcb28b7"
-  integrity sha512-0+uijmzYcnhC0hStDjm/cl2VYdrmVVBpe7Q8k9YBojxmR5tG8mvR9/nooQq3QSXiQqORDVOTY3XqMEqJVIzkHA==
+pkg-fetch@3.5.2:
+  version "3.5.2"
+  resolved "https://registry.yarnpkg.com/pkg-fetch/-/pkg-fetch-3.5.2.tgz#33e0d4f3b0c371fa9c133a10a081d5955d104376"
+  integrity sha512-KlRF3cDS4J5PRTKh5dkF5s+CYKa+4eUXzymWqTKPU/p3WmYlWZu7AS0dH8moPxg+QcJNB4/wu1wVO2a0Asv2Dw==
   dependencies:
     chalk "^4.1.2"
     fs-extra "^9.1.0"


emmansun:
Ignore pnpm test for node14  like https://github.com/vercel/pkg/pull/1613 or downgrade pnpm ?

https://pnpm.io/installation#compatibility

robertsLando:
Commented on the other PR regarding the pnpm issue

emmansun:
> Commented on the other PR regarding the pnpm issue

unless we are able to install older version of pnpm for nodejs older version.

robertsLando:
I don't see any problem with that, mine was just a question, giving that it's just a test util we could consider to just keep last pnpm version supported. I think you could re-open your PR

justajwolf:
> I don't see any problem with that, mine was just a question, giving that it's just a test util we could consider to just keep last pnpm version supported. I think you could re-open your PR

for pkg-fetch version, i think pkg should use version range: "~3.5.2" in deps. to avoid frequently update.

baparham:
> > I don't see any problem with that, mine was just a question, giving that it's just a test util we could consider to just keep last pnpm version supported. I think you could re-open your PR
> 
> for pkg-fetch version, i think pkg should use version range: "~3.5.2" in deps. to avoid frequently update. you think?

are you suggesting so that end users can let their package managers resolve to more recent yet-to-be released versions of `pkg-fetch` without needing to roll a new release of `pkg`?

I don't know the historical reasoning behind why this is hard-pinned to a version. Considering these packages are siblings, I find the risk to be low to allow for semver flexibility. If it were an external package, I would take a hard pass at it, since as we've seen, NOT hard pinning versions opens our end users up to supply chain attacks where one of the dependencies is hijacked and has a new patch version released that all of the sudden everyone starts using. But, like I said, the relationship between pkg and pkg-fetch is a bit closer and can afford some more flexibility considering that risk.

I'm open to it, but I also don't have all the info and would love to hear if there are any suggestions or arguments against loosening the version string.

robertsLando:
@baparham I think problem is that when we bump a pkg-fetch version there could be new nodejs binaries created for it so a new install of pkg will use those versions instead of the previous ones, we already spoke about this in another issue and could be a problem or not, if there are issues with latest nodejs binaries this could be a problem as you can have 2 installations of pkg one that is working and the other not and you don't understand why

baparham:
@robertsLando true, it does open users up to unknowingly getting newer versions of pkg-fetch in some circumstances, however, isn't this what lock files are meant to prevent (i.e. yarn.lock or package-lock.json)? Presumably users who forego the protections of  a lock file understand the consequences of dependency version volatility. Are there cases where a user might still get a new version of pkg-fetch even if they have a lock file maybe?

robertsLando:
> Are there cases where a user might still get a new version of pkg-fetch even if they have a lock file maybe

I don't think so

justajwolf:
> @baparham I think problem is that when we bump a pkg-fetch version there could be new nodejs binaries created for it so a new install of pkg will use those versions instead of the previous ones, we already spoke about this in another issue and could be a problem or not, if there are issues with latest nodejs binaries this could be a problem as you can have 2 installations of pkg one that is working and the other not and you don't understand why

Me,too. I also approve of this.

justajwolf:
> @robertsLando true, it does open users up to unknowingly getting newer versions of pkg-fetch in some circumstances, however, isn't this what lock files are meant to prevent (i.e. yarn.lock or package-lock.json)? Presumably users who forego the protections of a lock file understand the consequences of dependency version volatility. Are there cases where a user might still get a new version of pkg-fetch even if they have a lock file maybe?

I think that pkg-fetch is backward compatibility range patch versions. so pkg deps should use semver version limit major and minor, but hard-pinned to a version. you think?

Besides this，i think that pkg must be relate to pkg-fetch. pkg-fetch is sub package of the pkg.

baparham:
please rebase, node 14 test should be fixed on main now

baparham:
BTW, I think the decision will be to keep the hard pinned version (i.e. `"3.5.2"`), as this PR has it. When we release `pkg-fetch` it makes just as much sense to release an associated `pkg` to keep them in lock-step.

justajwolf:
> 



> BTW, I think the decision will be to keep the hard pinned version (i.e. `"3.5.2"`), as this PR has it. When we release `pkg-fetch` it makes just as much sense to release an associated `pkg` to keep them in lock-step.

ok, i approve this decision.  when pkg-fetch version to update, pkg version should also update version?

emmansun (approved):
rebase after node14 test fix. @baparham, pls kindly help to review it.

robertsLando (approved):

emmansun:
Hi @robertsLando , pls merge this PR, thx!

PR #1915, opened 2023-04-20
get fs/Module/childProcess/process native method.

xlch88:
I think definitely someone needs this. This solves a lot of problems.


baparham:
Not humoring PRs that dont' follow common etiquette and don't explain the purpose behind the changes.

PR #1916, opened 2023-04-20
fix: fs/Module/childProcess/process native method.

xlch88:
Simply using `PKG_EXECPATH` to distinguish whether it is a child process is unreliable, which may cause many problems or side effects. For example, when using `child_process.exec` to execute itself, the first parameter must be passed into the js file path. That is to say, when other programs that use pkg to package execute themselves, the first parameter must be passed in the js file path.

The maintainer should consider refactoring this part or directly provide the native method of nodejs. In fact, the maintainer has "backed up" the native method of nodejs
```javascript
const ancestor = {
    spawn: childProcess.spawn,
    spawnSync: childProcess.spawnSync,
    execFile: childProcess.execFile,
    execFileSync: childProcess.execFileSync,
    exec: childProcess.exec,
    execSync: childProcess. execSync,
};
```
Now just add `childProcess._ancestor = ancestor;` on it and everything should be fine.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -773,6 +773,8 @@ function payloadFileSync(pointer) {
       return fd;
     }
   }
+  
+  fs._ancestor = ancestor;
 
   fs.createReadStream = function createReadStream(path_) {
     if (!insideSnapshot(path_)) {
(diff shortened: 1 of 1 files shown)


baparham:
two things:
1. commit message is still unacceptable
2. I don't see a huge need for this, no one else has raised this issue or bumped it up or +1'd it or anything, and it seems too obscure to include. I can't wrap my head around the security implications of exposing the native node functionality well enough to give this a thumbs up and merge it in so I won't be approving this.
 
Furthermore, I don't think that pkg needs to keep adding features and enhancements, since we can't even get our act together to make regular releases (this one's on me, I'll be the first to admit).

A reasonable workaround for someone who actually needs this is to patch their own version of pkg and use that to build their own binaries.

PR #1923, opened 2023-04-26
fix: Add missing functions from restored fs.Stats

phated:
In my project, I noticed that a library that used `isCharacterDevice` on a file from the snapshot crashed. This adds the missing functions as noops (and I made isFIFO use noop so it was clearer).

I thought about making these mirror their files that were included, but I don't think either makes sense within the virtual fs. Please correct me if I'm wrong.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -1419,6 +1419,8 @@ function payloadFileSync(pointer) {
     delete s.isSocketValue;
     delete s.isSymbolicLinkValue;
 
+    s.isBlockDevice = noop;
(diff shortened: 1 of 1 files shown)


robertsLando:
@phated wouldn't it make more sense to return `false` instead of using noop there? Otherwise simply create a function `returnFalse` and use that there?

phated:
@robertsLando that's what noop does. It's used for Dirent prototype too

baparham (approved):
I found the noop function up near line 1221. LGTM 👍

Issue #1861, opened 2023-02-04
MODULE_NOT_FOUND error but module exists


MasumNishat:
### What version of pkg are you using?

5.8.0

### What version of Node.js are you using?

16.13.0

### What operating system are you using?

Windows 11

### What CPU architecture are you using?

x86_64

### What Node versions, OSs and CPU architectures are you building for?

default

### Describe the Bug

```cmd
pkg/prelude/bootstrap.js:1949
      if (error.code !== 'MODULE_NOT_FOUND') throw error;
                                             ^

TypeError: String.prototype.startsWith called on null or undefined
    at startsWith (<anonymous>)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:849:7)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1947:44)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.runMain (pkg/prelude/bootstrap.js:1983:12)
    at node:internal/main/run_main_module:17:47
```

### Expected Behavior

Should execute `app.exe`

### To Reproduce

1.    Create `app.js`
2.    Add:
```js
const child_process = require('child_process');
child_process.spawn('cmd.exe', ['/k', 'start cmd'], {
    detached: true,
    stdio: ['ignore', 'ignore', 'ignore']
}).unref();
(async () => {
    await new Promise(r => setTimeout(r, 10000))
    process.exit()
})()
```

3.    Compile with `pkg app.js --targets node16-win-x64`
4. Double-click on `app.exe`
5. Wait for the child window to appear
6. Type `app.exe` and press `Enter`

ryanrubleycoates:
We see this randomly across a fleet of 50K devices now and then, but restarting the process works eventually...

MasumNishat:
Is there any other process to restart self, I need it to add auto-update feature in my app. this code works fine before compiling.

lucasvieceli:
npm install string.prototype.startswith

MasumNishat:
I don't understand what you are telling @lucasvieceli . I'm trying to restart the program, not check some string starts with. You should first understand the problem. 

@ryanrubleycoates I don't know how you have tried. My code not working. Is there any mistake I have done here?
If not then You should mark it as an issue and fix it. else, you tell me that I'm wrong and instruct me on the system.
You are telling the restart method works but not telling exactly how it works. I think I'm very much clear about my problem.

lucasvieceli:
[...]

PR #1949, opened 2023-06-20
fix: missing entrypoint when launched from self-created child process

Luzzifus:
There is a bug that occurs when a packaged executable (Windows 11) is launched from a child process which was created by the executable itself (like for a self updating executable). This bug was reported already in #1861.

It seems that there is a conditional block at the start of `bootstrap.js` which has the potential to cause `ENTRYPOINT` to become `undefined`. I have to admit that I don't understand the purpose of the that specific code block, but I understand as much as that there always needs to be an entrypoint to successfully launch the executable. So I simply added an `else` statement as a fall back to the `DEFAULT_ENTRYPOINT`.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -80,14 +80,12 @@ if (process.env.PKG_EXECPATH === EXECPATH) {
   if (process.argv[1] && process.argv[1] !== '-') {
     // https://github.com/nodejs/node/blob/1a96d83a223ff9f05f7d942fb84440d323f7b596/lib/internal/bootstrap/node.js#L269
     process.argv[1] = path.resolve(process.argv[1]);
-  } else {
-    process.argv[1] = DEFAULT_ENTRYPOINT;
   }
 } else {
   process.argv[1] = DEFAULT_ENTRYPOINT;
 }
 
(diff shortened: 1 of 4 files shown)


robertsLando (dismissed):
Could you add a test that covers this?

Luzzifus:
> Could you add a test that covers this?

I added one. I also changed the actual fix a little bit to make sure it really only is a fallback for the case that `argv[1]` is `undefined` after all other modifications to `argv`.

robertsLando (commented):
Thanks for the test. Could you also fix lint issues please?

Luzzifus:
> Thanks for the test. Could you also fix lint issues please?

How would I do that? I ran `yarn lint` and `yarn fix` locally and it didn't change anything and I also got neither warnings nor errors for the files I changed/added.

robertsLando:
The error is shown here: https://github.com/vercel/pkg/actions/runs/5321783982/jobs/9638103625?pr=1949#step:6:19

Try using `npx eslint lib prelude test --fix`

Luzzifus:
Got it. Fixed.

robertsLando (approved):

PR #1996, opened 2023-09-26
fix: diagnostic folder size and humanize size

robertsLando:
- Fix issue of folder size calculation
- Print sizes in human friendly format
- Print folders/files that exceed a specific size. This can be set using `SIZE_LIMIT_PKG` and `FOLDER_LIMIT_PKG` env vars

### DIFF
--- a/prelude/diagnostic.js
+++ b/prelude/diagnostic.js
@@ -25,11 +25,9 @@ function humanSize(bytes) {
   const path = require('path');
   const win32 = process.platform === 'win32';
 
-  // set a size limit of 5MB
   const sizeLimit = process.env.SIZE_LIMIT_PKG
     ? parseInt(process.env.SIZE_LIMIT_PKG, 10)
     : 5 * 1024 * 1024;
-  // set a size limit of 10MB
   const folderLimit = process.env.FOLDER_LIMIT_PKG
     ? parseInt(process.env.FOLDER_LIMIT_PKG, 10)
     : 10 * 1024 * 1024;
--- a/prelude/diagnostic.js
+++ b/prelude/diagnostic.js
@@ -4,14 +4,42 @@
 
 'use strict';
 
+function humanSize(bytes) {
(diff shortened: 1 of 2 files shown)


prelude/diagnostic.js
```diff
@@ -4,14 +4,42 @@
...
+  if (bytes === 0) {
+    return 'n/a';
+  }
+
+  const i = Math.floor(Math.log(bytes) / Math.log(1024));
+
+  if (i === 0) {
+    return `${bytes} ${sizes[i]}`;
+  }
+
```
erossignon, line 19: Strictly speaking we should add a boundary checking here.
If someone passed a number of bytes that's larger than what 'TB' can represent, the function would try to access an index in the sizes array that doesn't exist, which would result in undefined.
But this is unlikely that someone will try to pack a file that is greater than a TB.

prelude/diagnostic.js
```diff
@@ -4,14 +4,42 @@
...
+  return `${(bytes / 1024 ** i).toFixed(1)} ${sizes[i]}`;
+}
+
 (function installDiagnostic() {
   const fs = require('fs');
   const path = require('path');
   const win32 = process.platform === 'win32';
 
+  // set a size limit of 5MB
```
erossignon, line 28: The comment is lying.

erossignon (approved):

prelude/diagnostic.js
robertsLando: > But this is unlikely that someone will try to pack a file that is greater than a TB.

Yeah I wouldn't expext that BTW js would just return `undefined` there, so something that would throw anyway

prelude/diagnostic.js
robertsLando: Agree, removed both comments

robertsLando:
@erossignon FYI giving that I don't receive any feedback from Vercel anymore, I created my own forks of both pkg and pkg-fetch:

https://github.com/yao-pkg/pkg-fetch
https://github.com/yao-pkg/pkg

Issue #1778, opened 2022-10-10
removeTemporaryFolderAndContent is failing, ENOENT: no such file or directory


jgoux:
### What version of pkg are you using?

5.8.0

### What version of Node.js are you using?

16.16.0

### What operating system are you using?

macOS

### What CPU architecture are you using?

Apple M1

### What Node versions, OSs and CPU architectures are you building for?

default

### Describe the Bug

When the program exit and involves a native node dependency (in my case `better-sqlite3`), it seems like it's trying to delete a temporary folder that doesn't exist (or was already deleted?).

Note that to make `better-sqlite3` works I had to add it as an asset with this config:

```json
{
  "pkg": {
    "assets": [
      "./node_modules/better-sqlite3/build/Release/better_sqlite3.node"
    ]
  }
}
```

This exact line seems to fail: https://github.com/vercel/pkg/blob/614c02a06830e0a4a898e6cf4eb35dbbf6110af3/prelude/bootstrap.js#L710

Here is the error:

```bash
Error: ENOENT: no such file or directory, stat '/var/folders/b1/0zl3wj_s47j_9fyf7djk6c5h0000gn/T/pkg-tuW0D3'
    at Object.statSync (node:fs:1551:3)
    at Object.statSync (pkg/prelude/bootstrap.js:1479:32)
    at __node_internal_ (node:internal/fs/utils:799:8)
    at Object.rmSync (node:fs:1226:13)
    at removeTemporaryFolderAndContent (pkg/prelude/bootstrap.js:704:10)
    at process.<anonymous> (pkg/prelude/bootstrap.js:711:5)
    at process.emit (node:events:539:35)
    at process.emit (node:domain:475:12)
```

Note that the error only happens once, then it seems like as the folders are present in `tmp`, it works.

Before running the program for the first time (the program is failing with the above error):

```bash
$ ls /var/folders/b1/0zl3wj_s47j_9fyf7djk6c5h0000gn/T/pkg
ls: /var/folders/b1/0zl3wj_s47j_9fyf7djk6c5h0000gn/T/pkg: No such file or directory
```

After running the program once (the program works without the error):

```bash
$ ls /var/folders/b1/0zl3wj_s47j_9fyf7djk6c5h0000gn/T/pkg
108dae79226db6e73b1951923db8a79c5e736daeecda33b14b29d7900759e235
[...]

PR #2026, opened 2023-11-27
Do not remove temporary folder if it doesn't exist

sebbo2002:
When executing binaries created with pkg, I also occasionally encountered the error described in #1778, for example:

```
Error: ENOENT: no such file or directory, stat '/var/folders/q3/7_wg79yj5mg14320hlz4sqr00000gn/T/pkg-bnLwAL'
    at Object.statSync (node:fs:1596:3)
    at Object.statSync (pkg/prelude/bootstrap.js:1476:32)
    at __node_internal_ (node:internal/fs/utils:801:8)
    at Object.rmSync (node:fs:1271:13)
    at removeTemporaryFolderAndContent (pkg/prelude/bootstrap.js:704:10)
    at process.<anonymous> (pkg/prelude/bootstrap.js:711:5)
    at process.emit (node:events:549:35)
    at process.emit (node:domain:482:12)
```

Unfortunately irregular and not reproducible. I would therefore like to add this line so that the temporary folder is not deleted in the first place if it does not exist.

### DIFF
--- a/prelude/bootstrap.js
+++ b/prelude/bootstrap.js
@@ -688,7 +688,7 @@ function payloadFileSync(pointer) {
   // ///////////////////////////////////////////////////////////////
 
   function removeTemporaryFolderAndContent(folder) {
-    if (!folder) return;
+    if (!folder || !fs.existsSync(folder)) return;
     if (NODE_VERSION_MAJOR <= 14) {
       if (NODE_VERSION_MAJOR <= 10) {
         // folder must be empty


austen-herbst (approved):

austen-herbst:
@leerob Would it be possible to get this change in a release? We are facing the same issue as @sebbo2002 (described in #1778) and would like to take advantage of this fix.

